PR #202, opened 2024-10-18
Fixes to support newer pinecone sdk

manushreegangwar:
This PR is to add support for newer pinecone SDK: https://docs.pinecone.io/reference/python-sdk

### DIFF
--- a/fiftyone/brain/internal/core/pinecone.py
+++ b/fiftyone/brain/internal/core/pinecone.py
@@ -250,20 +250,14 @@ class PineconeSimilarityIndex(SimilarityIndex):
             if self.config.index_type in [None, "serverless"]:
                 from pinecone import ServerlessSpec
 
-                cloud = self.config.cloud if self.config.cloud else "aws"
-                environment = (
-                    self.config.environment
-                    if self.config.environment
-                    else "us-east-1"
-                )
                 if not self._pinecone.has_index(self.config.index_name):
                     self._pinecone.create_index(
                         name=self.config.index_name,
                         dimension=dimension,
                         metric=metric,
                         spec=ServerlessSpec(
-                            cloud=cloud,
-                            region=environment,
+                            cloud=self.config.cloud,
+                            region=self.config.environment,
                         ),
                     )
             elif self.config.index_type == "pod":
--- a/fiftyone/brain/internal/core/pinecone.py
+++ b/fiftyone/brain/internal/core/pinecone.py
@@ -40,21 +40,27 @@ class PineconeSimilarityConfig(SimilarityConfig):
         supports_prompts (None): whether this run supports prompt queries
         index_name (None): the name of a Pinecone index to use or create. If
             none is provided, a new index will be created
-        index_type (None): the index type to use when creating a new index
+        index_type (None): the index type to use when creating a new index.
+            The supported values are ``["serverless", "pod"]`` and the default
+            is ``"serverless"``
         namespace (None): a namespace under which to store vectors added to the
             index
         metric (None): the embedding distance metric to use when creating a
             new index. Supported values are
             ``("cosine", "dotproduct", "euclidean")``
         replicas (None): an optional number of replicas when creating a new
+            pod-based index
+        shards (None): an optional number of shards when creating a new
+            pod-based index
+        pods (None): an optional number of pods when creating a new pod-based
             index
-        shards (None): an optional number of shards when creating a new index
-        pods (None): an optional number of pods when creating a new index
-        pod_type (None): an optional pod type when creating a new index
+        pod_type (None): an optional pod type when creating a new pod-based
+            index
         api_key (None): a Pinecone API key to use
-        project_name (None): a Pinecone project to use
-        cloud (None): Cloud where Pinecone index would be hosted
-        environment (None): a Pinecone environment to use
+        cloud (None): a cloud to use when creating serverless indexes
+        region (None): a region to use when creating serverless indexes
+        environment (None): an environment to use when creating pod-based
+            indexes
     """
 
     def __init__(
@@ -72,8 +78,8 @@ class PineconeSimilarityConfig(SimilarityConfig):
         pods=None,
         pod_type=None,
         api_key=None,
-        project_name=None,
         cloud=None,
+        region=None,
         environment=None,
         **kwargs,
     ):
@@ -103,8 +109,8 @@ class PineconeSimilarityConfig(SimilarityConfig):
         # store privately so these aren't serialized
         self._api_key = api_key
         self._cloud = cloud
+        self._region = region
         self._environment = environment
-        self._project_name = project_name
 
     @property
     def method(self):
@@ -119,24 +125,28 @@ class PineconeSimilarityConfig(SimilarityConfig):
         self._api_key = value
 
     @property
-    def environment(self):
-        return self._environment
-
-    @property
     def cloud(self):
         return self._cloud
+
+    @cloud.setter
+    def cloud(self, value):
+        self._cloud = value
+
+    @property
+    def region(self):
+        return self._region
+
+    @region.setter
+    def region(self, value):
+        self._region = value
+
+    @property
+    def environment(self):
+        return self._environment
 
     @environment.setter
     def environment(self, value):
         self._environment = value
-
-    @property
-    def project_name(self):
-        return self._project_name
-
-    @project_name.setter
-    def project_name(self, value):
-        self._project_name = value
 
     @property
     def max_k(self):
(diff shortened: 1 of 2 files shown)


tests/intensive/test_similarity.py
```diff
@@ -235,13 +237,13 @@ def test_image_similarity_backends():
...
-        dataset.clear_cache()
+        # del index2
+        # dataset.clear_cache()
 
-        print(dataset.get_brain_info(brain_key))
+        # print(dataset.get_brain_info(brain_key))
 
-        index2 = dataset.load_brain_results(brain_key)
-        assert index2.total_index_size == 200
+        # index2 = dataset.load_brain_results(brain_key)
```
manushreegangwar, line 245: @brimoor Do you know understand `index2._index = None` when using `load_brain_results`?

manushreegangwar: This happens when we call `dataset.clean_cache` before `load_brain_results`. 

I have changed the order as a workaround. Let me know if you have a cleaner solution.

brimoor (changes requested):

fiftyone/brain/internal/core/pinecone.py
```diff
@@ -31,6 +32,8 @@ class PineconeSimilarityConfig(SimilarityConfig):
     """Configuration for the Pinecone similarity backend.
 
     Args:
+        cloud ("aws"): Cloud where Pinecone index would be hosted
+        environment ("us-east-1"): a Pinecone environment to use
```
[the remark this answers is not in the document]
manushreegangwar: https://docs.pinecone.io/reference/api/2024-07/control-plane/create_index

These are required parameters and pinecone doesn't have any defaults. I am setting these as the default values which are overwritten at config initialization so I wouldn't consider this hard-coding. I can change the implementation to handle this in `create_index`.

I will move them to the bottom of the list.

fiftyone/brain/internal/core/pinecone.py
```diff
@@ -183,17 +191,29 @@ class PineconeSimilarityIndex(SimilarityIndex):
...
+
+            self._pinecone = Pinecone(
+                api_key=self.config.api_key,
+            )
 
         try:
-            index_names = pinecone.list_indexes()
+            index_names = (
+                self._pinecone.list_indexes()
+                if self._pinecone
```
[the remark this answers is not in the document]
manushreegangwar: This will cause issues in `create_index`

fiftyone/brain/internal/core/pinecone.py
```diff
@@ -209,30 +229,76 @@ def _initialize(self):
...
+                        pod_type=pod_type,
+                        pods=pods,
+                        replicas=replicas,
+                        shards=shards,
+                    ),
+                )
+            while not self._pinecone.describe_index(
+                self.config.index_name
+            ).status["ready"]:
+                time.sleep(1)
```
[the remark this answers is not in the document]
manushreegangwar: Yes, `pinecone` takes times to initialize and I was running into issues because of that. Agree with the no timeout in core but I didn't find a method in pinecone SDK for this. I will check again to see if there's an alternative

tests/intensive/test_similarity.py
```diff
@@ -164,6 +164,8 @@ def test_image_similarity_backends():
 
     # sklearn backend
     ###########################################################################
+    if "clip_sklearn" in dataset.list_brain_runs():
```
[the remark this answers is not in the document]
manushreegangwar: These changes were made to handle the tests not reaching `dataset.delete()`. I personally don't like implicit assumption when running tests. I didn't know about `drop_existing_dataset=True`. Thanks for the suggestion. I will try it out.

fiftyone/brain/internal/core/pinecone.py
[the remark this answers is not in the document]
brimoor: The MongoDB backend has a similar behavior so we added a `ready` property:
https://github.com/voxel51/fiftyone-brain/blob/2374ef5b20985ad65b38665d55effd1cce725a7a/fiftyone/brain/internal/core/mongodb.py#L282-L301

That can be used as documented here:
https://docs.voxel51.com/integrations/mongodb.html#checking-if-an-index-is-ready

Can you adopt that pattern here for consistency?

tests/intensive/test_similarity.py
brimoor: There shouldn't need to be any reordering of commands in the tests; that indicates that the updated PR doesn't quite behave the same as the previous implementation did.

The intent of `dataset.load_brain_results(brain_key)` is that it establishes a connection to the existing Pinecone index if one was previously created. Take a look at the implementation and see if you can spot why the behavior changed from the previous implementation where the tests were passing.

brimoor: To clarify: if you call `dataset.load_brain_results(brain_key)` in the same Python session where you originally called `compute_similarity(dataset, brain_key=brain_key, ...)`, it will return the cached object, which has already taken care of initializing `_index` (indeed, it created the index).

However, each `SimilarityIndex` class also needs to support the workflow where a user wants to connect to an existing similarity index on an existing dataset in a fresh Python session:

```py
import fiftyone as fo

dataset = fo.load_dataset(...)
index = dataset.load_brain_results(brain_key)

assert index._index is not None
```

manushreegangwar: Yes, `dataset.clean_cache` destroys `index._index` for pinecone and `load_brain_results` doesn't load it. There's no `index._index` in sklearn backend which is why the test passes for sklearn.

I believe this behavior precedes the PR. I will dig around to see how to fix the issue.

manushreegangwar:
Thanks for the feedback, @brimoor. I have addressed your comments and  all "pinecone" related tests in test_similarity.py are passing. Ready for review :mag:

fiftyone/brain/internal/core/pinecone.py
```diff
@@ -183,17 +191,29 @@ class PineconeSimilarityIndex(SimilarityIndex):
...
+        self._pinecone = None
         self._initialize()
 
     def _initialize(self):
-        pinecone.init(
-            api_key=self.config.api_key,
-            environment=self.config.environment,
-            project_name=self.config.project_name,
-        )
+        try:
```
brimoor, line 198: I believe that `pinecone.init()` was removed in pinecone v3. That's multiple versions ago, so after a bit more thought, I think it would be best to just require:

```py
    def ensure_usage_requirements(self):
        fou.ensure_package("pinecone-client>=3")
```

so that we can remove the `if` statements in this implementation and keep it clean. Sound good to you?

brimoor: Please double check that `pinecone-client>=3` is in fact correct 😄

fiftyone/brain/internal/core/pinecone.py
```diff
@@ -243,9 +309,29 @@ def index(self):
...
 
+    @property
+    def ready(self):
+        if self._pinecone:
+            return self._pinecone.describe_index(
+                self.config.index_name
+            ).status["ready"]
+        return pinecone.describe_index(self.config.index_name).status["ready"]
+
+    def verify_total_index_size(self, expected_size, timeout=10, interval=1):
```
brimoor, line 322: It's unfortunate that Pinecone itself doesn't have a concept of what MongoDB calls "write concern" that allows you to wait for their product to reflect the changes that were requested 😢

In practice, when potentially multiple people are working on the same Pinecone index, it feels a bit hacky to wait for the index to reach a certain size. Maybe someone else deleted a sample from the index while this was polling, for example. The ideal thing would be "wait until Pinecone acknowledges that my current transaction completes". But that's not something we should build if Pinecone doesn't support it.

In my opinion, this feels more like a util function that we should put in the `tests` folder to start with so that our integration tests pass (it could be used by any backend, FWIW).

brimoor (changes requested):
Almost there! Great updates so far 💪

fiftyone/brain/internal/core/pinecone.py
brimoor: If you feel strongly that this is a util we should expose to users, then we can keep it here. Up to you 😄

fiftyone/brain/internal/core/pinecone.py
manushreegangwar: Done. Anything >= 3.2 should be fine. 

`pinecone-client==3.0` is a mystery as it neither has `init` not `Pinecone` :shrug:

fiftyone/brain/internal/core/pinecone.py
manushreegangwar: :+1: Moved this to `test_similarity.py`.

brimoor (changes requested):

fiftyone/brain/internal/core/pinecone.py
```diff
@@ -18,6 +18,7 @@
     SimilarityIndex,
 )
 import fiftyone.brain.internal.core.utils as fbu
+from pinecone import Pinecone, ServerlessSpec, PodSpec
```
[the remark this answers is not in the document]
manushreegangwar: I didn't notice `lazy_import`. Thanks, switched pinecone to use that.

brimoor (approved):
LGTM

PR #203, opened 2024-10-21
Feature/leaky splits

jacobsela:
Computes leaks on a dataset using embeddings and the similarity brain method.

To try out:

```py
import fiftyone as fo
import fiftyone.brain as fob
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("cifar10")

dataset.compute_embeddings(model, "embeds_clip", batch_size=32, num_workers=16)

index_clip, leaks_clip = fob.compute_leaky_splits(
    dataset,
    "foobar",
    split_field="split",
    threshold=0.2,
    embeddings_field="embeds_clip",
)
```


mwoodson1:
@jacobsela when finalized could you provide a code snippet for how you imagine this feature working?

fiftyone/brain/internal/core/leaky_splits.py
```diff
@@ -0,0 +1,373 @@
...
+    def _compute_hashes(self, samples):
+        for s in samples.iter_samples():
+            hash = str(self._hash_function(s["filepath"]))
+            self._hash2ids[hash].append(s["id"])
+            if self.config.hash_field:
+                s[self.config.hash_field] = hash
+                s.save()
+
+    @staticmethod
+    def _image_hash(image, hash_size=24):
```
mwoodson1, line 321: I think image hashing is a generally useful enough function that it should be pulled out. For example, the `compute_exact_duplicates` method uses [this](https://github.com/voxel51/fiftyone-brain/blob/2374ef5b20985ad65b38665d55effd1cce725a7a/fiftyone/brain/internal/core/duplicates.py#L93) hash but could also easily be changed to use a image hash.

jacobsela:
@mwoodson1 
Basic snippet I used in the demo:

```
import fiftyone as fo
import fiftyone.brain.internal.core.leaky_splits as ls

config = ls.LeakySplitsSKLConfig(
    split_tags=['train', 'test'],
    model="resnet18-imagenet-torch"
)

# skl backend
index = ls.LeakySplitsSKL(config).initialize(dataset, "foo")

index.set_threshold(0.1)
leaks = index.leaks

session = fo.launch_app(leaks, auto=False)

# hash backend
config = ls.LeakySplitsHashConfig(
    split_tags=['train', 'test'],
    method='image',
    hash_field='hash'
)

index = ls.LeakySplitsHash(config).initialize(dataset, "foo")

session = fo.launch_app(index.leaks, auto=False)
```

jacobsela:
@mwoodson1 Thanks for the feedback, I agree that this isn't ideal. I'm holding off on creating the final `compute_leaks` (or `compute_leaky_splits` as it currently is in the code) until we finalize what we want the behavior to look like (e.g. in terms of thresholds). Putting together a final easy to use function at the end should be quick so I'd rather do it once.

fiftyone/brain/internal/core/leaky_splits.py
```diff
@@ -0,0 +1,497 @@
...
+
+class LeakySplitsSKL(skl_sim.SklearnSimilarity):
+    def initialize(self, samples, brain_key):
+        return LeakySplitsSKLIndex(
+            samples, self.config, brain_key, backend=self
+        )
+
+
+class LeakySplitsSKLIndex(
+    skl_sim.SklearnSimilarityIndex, LeakySplitIndexInterface
```
brimoor, line 312: I'm not sure that subclassing from `SklearnSimilarityIndex` makes sense here. It has the side effect of causing a full copy of all of the embeddings to be stored on the dataset when `save_run_results()` is called in the main method.

I would instead argue that "leaky splits" is an algorithm that *may use* embeddings, but it shouldn't be directly responsible for storing a copy of the embeddings. Those should only be stored exactly once by `compute_similarity()`.

Downstream methods like this should be able to use pre-computed embeddings. Or, if no embeddings are provided, then this method can compute its own embeddings, but they won't actually be stored permanently. 

For context, this is how `compute_uniqueness()` and `compute_representativeness()` and `compute_visualization()` all work.

fiftyone/brain/internal/core/leaky_splits.py
```diff
@@ -0,0 +1,497 @@
...
+import fiftyone.core.validation as fov
+import fiftyone.zoo as foz
+
+fbu = fou.lazy_import("fiftyone.brain.internal.core.utils")
+
+_DEFAULT_MODEL = "clip-vit-base32-torch"
+_DEFAULT_BATCH_SIZE = None
+
+
+def compute_leaky_splits(
```
brimoor, line 31: I agree with @mwoodson1 that, in order to ship this as part of the Brain, it should be exposed at the top-level via `fob.compute_leaky_splits()` or similar.

fiftyone/brain/internal/core/leaky_splits.py
```diff
@@ -0,0 +1,497 @@
...
+    split_views=None,
+    split_field=None,
+    split_tags=None,
+    threshold=0.2,
+    embeddings_field=None,
+    model=None,
+    model_kwargs=None,
+    patches_field=None,
+    metric="cosine",
+    **kwargs,
```
brimoor, line 43: There's no way to chose the hash-based backend right? That seems missing

fiftyone/brain/internal/core/leaky_splits.py
```diff
@@ -0,0 +1,497 @@
...
+        embeddings_field (None): field for embeddings to feed the index. This argument's
+            behavior depends on whether a ``model`` is provided, as described
+            below.
+
+            If no ``model`` is provided, this argument specifies the field of pre-computed
+            embeddings to use:
+
+            If a ``model`` is provided, this argument specifies where to store
+            the model's embeddings:
+        brain_key (None): a brain key under which to store the results of this
```
brimoor, line 88: Given my comments below about how I don't think this method should store its own embeddings, I'm not sure it needs a `brain_key`?

A `brain_key` is only necessary/useful if a method permanently stores data on the samples or the dataset itself.

For context, `fob.compute_exact_duplicates()` does not use a brain key because it doesn't store anything; it just returns a moment in time snapshot of the duplicates that currently exist on the dataset. It is up to the user to do what they will with that information.

fiftyone/brain/internal/core/leaky_splits.py
jacobmarks: @brimoor I mostly agree with this, but I think the default behavior of not storing the embeddings when the algorithm is run is potentially very frustrating for users, who need to recompute millions of embeddings.

fiftyone/brain/internal/core/leaky_splits.py
jacobmarks: I agree, it needs to follow the pattern `fob.compute_X()` that other Brain methods do. Preference for calling it `compute_leaks()`, rather than `compute_leaky_splits()`

fiftyone/brain/internal/core/leaky_splits.py
jacobmarks: Also, this method would not ever interface with/act on patches, right? Can safely remove that from explicitly in function signatures

fiftyone/brain/internal/core/leaky_splits.py
jacobmarks: Not a strongly held opinoin, but I disagree. The "leaks" will depend on the hashing method, and on the view you are computing on. I would expect results to differ when either or both is changed. One could absolutely be interested in comparing two runs.

fiftyone/brain/internal/core/leaky_splits.py
```diff
@@ -0,0 +1,497 @@
...
+        self._dataset = samples._dataset
+        self._compute_hashes(samples)
+
+    @property
+    def _hash_function(self):
+        if self.config.method == "filepath":
+            return fou.compute_filehash
+
+        elif self.config.method == "image":
+            return LeakySplitsHashIndex._image_hash
```
jacobmarks, line 434: I could be missing it, but I don't see how `hash_size` propagates into `LeakySplitsHashIndex` to override the default

fiftyone/brain/internal/core/leaky_splits.py
```diff
@@ -0,0 +1,497 @@
...
+        diff = resized[:, 1:] > resized[:, :-1]
+
+        # Convert the difference image to a binary hash
+        # hash_value = sum([2 ** i for (i, v) in enumerate(diff.flatten()) if v])
+
+        # Convert the difference image to a binary hash
+        binary_string = "".join(["1" if v else "0" for v in diff.flatten()])
+
+        # Convert the binary string to a hexadecimal string
+        hex_hash = f"{int(binary_string, 2):0{hash_size * hash_size // 4}x}"
```
jacobmarks, line 473: Link/cite the source for this math/implementation

jacobmarks (changes requested):
Echoing Brian's comments, plus a few questions on design decisions, parameters, etc

fiftyone/brain/internal/core/leaky_splits.py
jacobsela: This is just a matter of exposing this function in the `__init__`, correct?

fiftyone/brain/internal/core/leaky_splits.py
jacobsela: @brimoor Hash based backend is in a separate class, `LeakySplitsHash`. @jacobmarks I'll remove the patches stuff.

fiftyone/brain/internal/core/leaky_splits.py
brimoor: Yeah I get it. On the other hand, another source of annoyance would be having multiple copies of a potentially large number of embeddings duplicately stored on your dataset. And if you don't explicitly specify a model to use, you don't even know what the embeddings are as the default model used is not currently made known to the user.

The current mental model is basically:
- Provide a simple syntax `compute_uniqueness(dataset)` with minimal parameters that gives you the insight without concerning you with how it happens necessarily
- Provide advanced syntaxes `compute_uniqueness(dataset, **kwargs)` that allow for reusing embeddings, customizing the model used, etc.

fiftyone/brain/internal/core/leaky_splits.py
brimoor: I see that it is a separate class, but how do you actually tell FOB you want to use that method when you call `compute_leaky_splits()`?

fiftyone/brain/internal/core/leaky_splits.py
brimoor: My confusion is in how this concretely looks. Suppose I ran two leaky split methods and closed my FO session. What does it look to start a new session, call `load_dataset()` and `load_brain_results()` twice and start comparing my previous results? Is anything actually cached? 

I could definitely be missing something, its just hard to reason about this without some example code showing how one works with the output of this method.

jacobsela: I tend to agree with Jacob here, I definitely noticed a difference in results between different embedding models. Let me know how to proceed here.

jacobmarks: We allow `compute_visualization()` to store this type of information, and then in the panel we can toggle between runs. Could easily see a panel for leakage that allows you to toggle between leakage runs, loading the appropriate views and allowing you to take action on the results of one of these runs

fiftyone/brain/internal/core/leaky_splits.py
```diff
@@ -0,0 +1,497 @@
...
+# pylint: disable=no-member
+import cv2
+
+import eta.core.utils as etau
+
+import fiftyone.core.brain as fob
+import fiftyone.brain.similarity as sim
+import fiftyone.brain.internal.core.sklearn as skl_sim
+import fiftyone.brain.internal.core.duplicates as dups
+import fiftyone.brain.internal.core.utils as fbu
```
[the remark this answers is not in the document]
jacobsela: Good catch, removing.

fiftyone/brain/internal/core/leaky_splits.py
```diff
@@ -0,0 +1,497 @@
...
+    split_tags=None,
+    threshold=0.2,
+    embeddings_field=None,
+    model=None,
+    model_kwargs=None,
+    patches_field=None,
+    metric="cosine",
+    **kwargs,
+):
+    """Uses embeddings to index the samples or their patches so that you can
```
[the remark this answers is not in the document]
jacobsela: Arguably yes. Copy-paste augmentations are becoming more common. Very high resolution satellite/drone/whatever footage is often chopped up into smaller pieces for individual samples. There could be cases where only a patch of an image correlates. That said, it is a niche use case and not tested, so I agree it's best to remove.

fiftyone/brain/internal/core/leaky_splits.py
```diff
@@ -0,0 +1,497 @@
...
+
+    brain_method.save_run_results(samples, brain_key, results)
+
+    return results, leaks
+
+
+### GENERAL
+
+
+class LeakySplitsConfigInterface(object):
```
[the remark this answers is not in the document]
jacobsela: Mixins are kind of like that. The reason I went for this design choice is so that I only have to avoid code duplication for the overlap between the similarity and hash backends. I couldn't really find a clean way to fit them both into one class. If you have a suggestion I'd be happy to refactor.

fiftyone/brain/internal/core/leaky_splits.py
```diff
@@ -0,0 +1,497 @@
...
+        raise ValueError(f"Only one of the split arguments must be given.")
+
+    if split_views:
+        return split_views
+
+    if split_field:
+        return _field_to_views(samples, split_field)
+
+    if split_tags:
+        return _tags_to_views(samples, split_tags)
```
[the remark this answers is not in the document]
jacobsela: Thank you :)

fiftyone/brain/internal/core/leaky_splits.py
jacobsela: `compute_leaky_splits` doesn't support the hash method. I can extend it to do so. The hash based method seems to not be very useful because it looks for exact duplicates, something already covered by find exact duplicates and something that most people already have checks in place for. If it's important for it to be in the main function I'll add it

fiftyone/brain/internal/core/leaky_splits.py
brimoor: It's fine to use an `SklearnSimilarityIndex` instance internally (which could be a pre-existing one that users pass in), I just don't think you should subclass from it, because that has side effects like causing the embeddings to be stored permanently in the database under the `brain_key`.

fiftyone/brain/internal/core/leaky_splits.py
brimoor: I don't disagree, I just need some example code in the PR description so I can see what this looks like and try it myself 😄

fiftyone/brain/internal/core/leaky_splits.py
jacobsela: It doesn't, it's internal. I think it may be best to remove it entirely, but I can add control for it. What do you think?

fiftyone/brain/internal/core/leaky_splits.py
brimoor: Either it is useful and should be available via `compute_leaky_splits()` or it is not useful and should just be deleted from the codebase IMO. Adding unused code is just more work/confusion to maintain in the future.

fiftyone/brain/internal/core/leaky_splits.py
jacobsela: This was a huge sticking point for me when designing this, I'm glad it came up. I avoided holding an instance of the similarity index (despite it being more general because it allows the user to define what similarity index they want to use as long as it supports the duplicates mixin) because having a brain method running in the background without the user's knowledge seemed strange (wouldn't the instance still show up if the user listed all brain methods?).

fiftyone/brain/internal/core/leaky_splits.py
jacobsela: I lean towards removing it. Please let me know what you think.

fiftyone/brain/internal/core/leaky_splits.py
jacobsela:
```
dataset.compute_embeddings(model, "embeds_clip", batch_size=32, num_workers=16)
index_clip, leaks_clip = ls.compute_leaky_splits(dataset, "foobar", split_field='split', threshold=0.2, embeddings_field='embeds_clip')
```

jacobsela: Not sure how to change the PR description after making it, sorry about that.

fiftyone/brain/internal/core/leaky_splits.py
```diff
@@ -0,0 +1,497 @@
...
+
+    views = []
+    for val in field_values:
+        view = samples.match(F(field) == val)
+        views.append(view)
+
+    return views
+
+
+def _tags_to_views(samples, tags):
```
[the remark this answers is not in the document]
jacobsela: Check added.

fiftyone/brain/internal/core/leaky_splits.py
jacobsela: After meeting with Jacob:
1. Don't subclass from similarity
2. Have two arguments, brain_key and similarity_brain_key
3. if neither is passed, nothing is registered
4. if similarity_brain_key is passed and exists, load that run, otherwise compute and register
5. if similarity_brain_key is not passed, create but don't register similarity index
6. brain_key says if/where to save leaky splits run

jacobsela: @brimoor can you please let us know what you think about the above refactor?

fiftyone/brain/internal/core/leaky_splits.py
jacobsela: Suggested path forward:
1. remove hashing from leaky-splits
2. Backlog: add new hash function (image hash) as an option for `compute_duplicates`

fiftyone/brain/internal/core/leaky_splits.py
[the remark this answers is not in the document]
jacobsela: Will no longer be an issue after [refactor](https://github.com/voxel51/fiftyone-brain/pull/203#discussion_r1849036659). Hash will be removed so no longer any need to consolidate common functionality in one spot.

fiftyone/brain/internal/core/leaky_splits.py
jacobsela: same as above

jacobsela:
@brimoor @jacobmarks Implemented the things we talked about.

Some remaining points that may need addressing:
1. Scalability - I currently do a check to see if a previously loaded similarity index has the needed samples in it. This check is not scalable. In order to get around this, new embeddings should be added to the index on the fly. This would also require changing a lot of internal values in the similarity index that wouldn't even recognize the sample to have existed up until that point e.g. self._samples. Doing this would require being careful about a lot of different things that could potentially break so I am not doing it right now. If it's very important I can do it tomorrow. Otherwise, I'll backlog it so I can get back to the evaluation work.
2. I changed the id2split mechanism to be ran once at the start so that we have O(1) lookup during runtime which is done multiple times per sample. This comes at the cost of O(n) memory. This seems to be non important to me but if it is please let me know.
3. I'm sometimes having errors when I try to register the brain run, e.g.
```
Failed to run cleanup() for the brain method run with key 'full_save': `fiftyone.core.stages.Limit` has not been registered in the document registry.
            Importing the document class automatically registers it, has it
            been imported?
```
Importing the document class did not help. Do you see an issue with how I'm saving the brain runs?

jacobmarks:
> @brimoor @jacobmarks Implemented the things we talked about.
> 
> Some remaining points that may need addressing:
> 
> 1. Scalability - I currently do a check to see if a previously loaded similarity index has the needed samples in it. This check is not scalable. In order to get around this, new embeddings should be added to the index on the fly. This would also require changing a lot of internal values in the similarity index that wouldn't even recognize the sample to have existed up until that point e.g. self._samples. Doing this would require being careful about a lot of different things that could potentially break so I am not doing it right now. If it's very important I can do it tomorrow. Otherwise, I'll backlog it so I can get back to the evaluation work.
> 2. I changed the id2split mechanism to be ran once at the start so that we have O(1) lookup during runtime which is done multiple times per sample. This comes at the cost of O(n) memory. This seems to be non important to me but if it is please let me know.
> 3. I'm sometimes having errors when I try to register the brain run, e.g.
> 
> ```
> Failed to run cleanup() for the brain method run with key 'full_save': `fiftyone.core.stages.Limit` has not been registered in the document registry.
>             Importing the document class automatically registers it, has it
>             been imported?
> ```
> 
> Importing the document class did not help. Do you see an issue with how I'm saving the brain runs?

Thanks @jacobsela , really appreciate the prompt work! About to dive into the code, but a few comments/questions:

1. I don't think this check is necessary or the way to go. If I remember correctly, runs store the context including the view on which they were run. Given that you are only looking at the full images (not labels, other metadata aside from embeddings), comparing the current view to the view stored with the similarity run should be sufficient. Does this make sense? 
2. `id2split` mapping looks good to me
3. I tracked down [register_run](https://github.com/voxel51/fiftyone/blob/7785ecd8a3cc2988c14b05f19f51e7abb8994712/fiftyone/core/runs.py#L298), but can't seem to find where a view stage would cause an issue. If you try registering a run on a `DatasetView` with any other view stages, does the same thing happen? i.e. is it localized to `Limit`, or a general trend?

jacobsela:
Hey @jacobmarks, thanks for the feedback.

Regarding point 1:
The issue I'm concerned about is someone running similarity on some subset of the dataset then trying to use that run for a leaky splits run with a different subset. Similarity does save the run but it doesn't have any checks for things outside of the original samples passed to it.

Regarding point 3:
This happens for multiple view stages. Here is a reproducible example with `take`:
```
import fiftyone as fo
import fiftyone.zoo as foz
from fiftyone.core.document import Document
from fiftyone.brain import compute_leaky_splits
# load dataset
try:
    dataset = foz.delete_zoo_dataset('quickstart')
except:
    print("no need to delete")
dataset = foz.load_zoo_dataset('quickstart')
# random test train
train_view = dataset.take(100)
test_view = dataset.exclude(train_view.values('id'))
split_views = {
    'train' : train_view,
    'test' : test_view
}
# test loading full save
for key in ['full_save', 'full_save_similarity']:
    if key in dataset.list_brain_runs():
        dataset.delete_brain_run(key)
index, leaks = compute_leaky_splits(
    dataset,
    brain_key="full_save",
    split_views=split_views,
    similarity_brain_key="full_save_similarity",
    embeddings_field="full_save_embeds",
    model="resnet18-imagenet-torch"
)

index = dataset.load_brain_results('full_save')
leaks = index.leaks
session = fo.Session(leaks, auto=False)
```

This gives the output:
```
Downloading dataset to '/home/jacobs/fiftyone/quickstart'
Downloading dataset...
 100% |████|  187.5Mb/187.5Mb [734.2ms elapsed, 0s remaining, 255.5Mb/s]      
Extracting dataset...
Parsing dataset metadata
Found 200 samples
Dataset info written to '/home/jacobs/fiftyone/quickstart/info.json'
Loading existing dataset 'quickstart'. To reload from disk, either delete the existing dataset or provide a custom `dataset_name` to use
Failed to run cleanup() for the brain method run with key 'full_save': `fiftyone.core.stages.Take` has not been registered in the document registry.
            Importing the document class automatically registers it, has it
            been imported?
 100% |█████████████████| 100/100 [878.3ms elapsed, 0s remaining, 113.9 samples/s]      
 100% |█████████████████| 100/100 [763.2ms elapsed, 0s remaining, 131.0 samples/s]      
Computing duplicate samples...
Generating index for 200 embeddings...
Index complete
Generating neighbors graph for 197 embeddings...
Index complete
Duplicates computation complete
 100% |█████████████████| 100/100 [1.2s elapsed, 0s remaining, 85.0 samples/s]         
 100% |█████████████████| 100/100 [975.0ms elapsed, 0s remaining, 102.6 samples/s]      
Connected to FiftyOne on port 5151 at localhost.
If you are not connecting to a remote session, you may need to start a new session and specify a port
```

jacobmarks:
> Hey @jacobmarks, thanks for the feedback.
> 
> Regarding point 1: The issue I'm concerned about is someone running similarity on some subset of the dataset then trying to use that run for a leaky splits run with a different subset. Similarity does save the run but it doesn't have any checks for things outside of the original samples passed to it.
> 
> Regarding point 3: This happens for multiple view stages. Here is a reproducible example with `take`:
> 
> ```
> import fiftyone as fo
> import fiftyone.zoo as foz
> from fiftyone.core.document import Document
> from fiftyone.brain import compute_leaky_splits
> # load dataset
> try:
>     dataset = foz.delete_zoo_dataset('quickstart')
> except:
>     print("no need to delete")
> dataset = foz.load_zoo_dataset('quickstart')
> # random test train
> train_view = dataset.take(100)
> test_view = dataset.exclude(train_view.values('id'))
> split_views = {
>     'train' : train_view,
>     'test' : test_view
> }
> # test loading full save
> for key in ['full_save', 'full_save_similarity']:
>     if key in dataset.list_brain_runs():
>         dataset.delete_brain_run(key)
> index, leaks = compute_leaky_splits(
>     dataset,
>     brain_key="full_save",
>     split_views=split_views,
>     similarity_brain_key="full_save_similarity",
>     embeddings_field="full_save_embeds",
>     model="resnet18-imagenet-torch"
> )
> 
> index = dataset.load_brain_results('full_save')
> leaks = index.leaks
> session = fo.Session(leaks, auto=False)
> ```
> 
> This gives the output:
> 
> ```
> Downloading dataset to '/home/jacobs/fiftyone/quickstart'
> Downloading dataset...
>  100% |████|  187.5Mb/187.5Mb [734.2ms elapsed, 0s remaining, 255.5Mb/s]      
> Extracting dataset...
> Parsing dataset metadata
> Found 200 samples
> Dataset info written to '/home/jacobs/fiftyone/quickstart/info.json'
> Loading existing dataset 'quickstart'. To reload from disk, either delete the existing dataset or provide a custom `dataset_name` to use
> Failed to run cleanup() for the brain method run with key 'full_save': `fiftyone.core.stages.Take` has not been registered in the document registry.
>             Importing the document class automatically registers it, has it
>             been imported?
>  100% |█████████████████| 100/100 [878.3ms elapsed, 0s remaining, 113.9 samples/s]      
>  100% |█████████████████| 100/100 [763.2ms elapsed, 0s remaining, 131.0 samples/s]      
> Computing duplicate samples...
> Generating index for 200 embeddings...
> Index complete
> Generating neighbors graph for 197 embeddings...
> Index complete
> Duplicates computation complete
>  100% |█████████████████| 100/100 [1.2s elapsed, 0s remaining, 85.0 samples/s]         
>  100% |█████████████████| 100/100 [975.0ms elapsed, 0s remaining, 102.6 samples/s]      
> Connected to FiftyOne on port 5151 at localhost.
> If you are not connecting to a remote session, you may need to start a new session and specify a port
> ```


This is what I mean: you can get the view that a run was, well, run on, and compare it to the view passed into the new method.
```py
view = dataset.take(10)
fob.compute_similarity(view, brain_key="sim10", model="clip-vit-base32-torch")
view1 = dataset.load_brain_view("sim10")
assert(view1 == view)
```

In general, they don't need to be the same view, but should have the same samples:

```py
s1 = set(view1.values("id"))
s2 = set(view2.values("id"))
assert(s1 == s2)
```

fiftyone/brain/__init__.py
```diff
@@ -703,3 +703,97 @@ def compute_exact_duplicates(
...
+            to the model's ``Config`` when a model name is provided
+        similarity_backend: string, the similarity backend to use. The supported values are
+            ``fiftyone.brain.brain_config.similarity_backends.keys()`` and the
+            default is
+            ``fiftyone.brain.brain_config.default_similarity_backend``
+        similarity_config_dict: dict, used to build the similarity backend. Arguments take
+            precedence over the values in the dict (e.g. model)
+
+    Returns:
+        a :class:`fiftyone.brain.internal.core.leaky_splits.LeakySplitsSKLIndex`,
```
jacobmarks, line 779: I don't see `LeakySplitsSKLIndex` anywhere else. Am I missing something?

fiftyone/brain/internal/core/leaky_splits.py
```diff
@@ -0,0 +1,435 @@
...
+
+import fiftyone as fo
+from fiftyone import ViewField as F
+
+# pylint: disable=no-member
+import cv2
+
+import eta.core.utils as etau
+
+import fiftyone as fo
```
jacobmarks, line 16: Why are you importing `fo` twice?

jacobsela:
@jacobmarks Yes this works and is exactly what I do in line 172. @brimoor says this implementation doesn't scale well.

fiftyone/brain/internal/core/leaky_splits.py
```diff
@@ -0,0 +1,435 @@
...
+        self.embeddings_field = embeddings_field
+        self.model = model
+        self.model_kwargs = model_kwargs
+        self.similarity_backend = similarity_backend
+        self.similarity_config_dict = similarity_config_dict
+        super().__init__(**kwargs)
+
+    @property
+    def type(self):
+        return "Data Leakage"
```
jacobmarks, line 122: For consistency with existing code, let's keep it lowercase one word: `leakage`?

fiftyone/brain/internal/core/leaky_splits.py
```diff
@@ -0,0 +1,435 @@
...
+
+    @property
+    def method(self):
+        return "Similarity"
```
jacobmarks, line 126: lowercase

fiftyone/brain/internal/core/leaky_splits.py
```diff
@@ -0,0 +1,435 @@
...
+            if (
+                self.config.similarity_brain_key
+                in self.samples._dataset.list_brain_runs()
+            ):
+                self._similarity_index = (
+                    self.samples._dataset.load_brain_results(
+                        self.config.similarity_brain_key, load_view=True
+                    )
+                )
+                if self._similarity_index is not None:
```
jacobmarks, line 165: is this condition ever met? If so, when?

fiftyone/brain/internal/core/leaky_splits.py
```diff
@@ -0,0 +1,435 @@
...
+                    index_found = True
+                # check if brain run view lines up with samples provided
+                # TODO: Brian says there is a better way of doing this, need to think of how
+                similarity_view = self.samples._dataset.load_brain_view(
+                    self.config.similarity_brain_key
+                )
+                if not set(samples.values("id")) == set(
```
jacobmarks, line 172: Saw your comment. Not sure how to speed this up. @brimoor any pointers?

fiftyone/brain/internal/core/leaky_splits.py
```diff
@@ -0,0 +1,435 @@
...
+                    raise ValueError(
+                        "Provided similarity run doesn't include all samples given. "
+                        "Please rerun the similarity with all of the wanted samples."
+                    )
+            else:
+                self._save_similarity_index = True
+
+        # create new similarity index
+        if not index_found:
+            sim_conf_dict_aux = {
```
jacobmarks, line 184: rename from aux to kwargs?

fiftyone/brain/internal/core/leaky_splits.py
```diff
@@ -0,0 +1,435 @@
...
+        self._leaks = duplicates
+
+        return duplicates
+
+    @property
+    def num_leaks(self):
+        """Returns the number of leaks found."""
+        return self.leaks.count
+
+    def leaks_by_sample(self, sample):
```
jacobmarks, line 269: I would expect something called `leaks_by_samples` to return a dictionary, not give the leaks for a single sample.

jacobmarks (changes requested):
Looking much better! I haven't tested out all of the edge cases yet, but this is coming along quite nicely!

fiftyone/brain/__init__.py
```diff
@@ -703,3 +703,97 @@ def compute_exact_duplicates(
...
+    threshold=0.2,
+    similarity_brain_key=None,
+    embeddings_field=None,
+    model=None,
+    model_kwargs=None,
+    similarity_backend=None,
+    similarity_config_dict=None,
+    **kwargs,
+):
+    """Uses embeddings to index the samples so that you can
```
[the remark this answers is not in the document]
jacobsela: Changing to "Uses a similarity index or creates one on the spot to find leaks."

fiftyone/brain/__init__.py
```diff
@@ -703,3 +703,97 @@ def compute_exact_duplicates(
...
+    find leaks.
+
+    Calling this method only creates the index. You can then call the methods
+    exposed on the retuned object to perform the following operations:
```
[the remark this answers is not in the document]
jacobsela: Good catch

fiftyone/brain/internal/core/leaky_splits.py
jacobsela: fixed

fiftyone/brain/internal/core/leaky_splits.py
jacobsela: was supposed to be in case of bad loading, I can remove it.

jacobsela: honestly I'm not sure what would be returned in that case anyways

fiftyone/brain/internal/core/leaky_splits.py
jacobsela: changed

fiftyone/brain/internal/core/leaky_splits.py
jacobsela: That may be useful functionality and something I can add. It's effectively what the main `leaks` property gives but in view form. What would you recommend changing this method's name to?

fiftyone/brain/internal/core/leaky_splits.py
jacobsela: changed

fiftyone/brain/internal/core/leaky_splits.py
jacobmarks: Maybe `leaks_for_sample()`?

fiftyone/brain/internal/core/leaky_splits.py
```diff
@@ -0,0 +1,382 @@
...
+    views = {}
+    for tag in tags:
+        view = samples.match_tags([tag])
+        if len(view) < 1:  # no samples in tag
+            raise ValueError(
+                f"One of the tags provided, '{tag}', has no samples. Make sure every tag has at least one sample."
+            )
+        views[tag] = view
+
+    for i, v in enumerate(views):
```
jacobmarks, line 364: Running into this error:
```
[341](https://file+.vscode-resource.vscode-cdn.net/Users/jacobmarks/Desktop/~/Desktop/work/brain/fiftyone-brain/fiftyone/brain/internal/core/leaky_splits.py:341) if split_tags:
--> [342](https://file+.vscode-resource.vscode-cdn.net/Users/jacobmarks/Desktop/~/Desktop/work/brain/fiftyone-brain/fiftyone/brain/internal/core/leaky_splits.py:342)     return _tags_to_views(samples, split_tags)

File ~/Desktop/work/brain/fiftyone-brain/fiftyone/brain/internal/core/leaky_splits.py:377, in _tags_to_views(samples, tags)
    [375](https://file+.vscode-resource.vscode-cdn.net/Users/jacobmarks/Desktop/~/Desktop/work/brain/fiftyone-brain/fiftyone/brain/internal/core/leaky_splits.py:375) for i, v in enumerate(views):
    [376](https://file+.vscode-resource.vscode-cdn.net/Users/jacobmarks/Desktop/~/Desktop/work/brain/fiftyone-brain/fiftyone/brain/internal/core/leaky_splits.py:376)     other_tags = [t for j, t in enumerate(tags) if not i == j]
--> [377](https://file+.vscode-resource.vscode-cdn.net/Users/jacobmarks/Desktop/~/Desktop/work/brain/fiftyone-brain/fiftyone/brain/internal/core/leaky_splits.py:377)     if len(v.match_tags(other_tags)) > 0:
    [378](https://file+.vscode-resource.vscode-cdn.net/Users/jacobmarks/Desktop/~/Desktop/work/brain/fiftyone-brain/fiftyone/brain/internal/core/leaky_splits.py:378)         raise ValueError(
    [379](https://file+.vscode-resource.vscode-cdn.net/Users/jacobmarks/Desktop/~/Desktop/work/brain/fiftyone-brain/fiftyone/brain/internal/core/leaky_splits.py:379)             f"One or more samples have more than one of the tags provided! Every sample should have at most one of the tags provided."
    [380](https://file+.vscode-resource.vscode-cdn.net/Users/jacobmarks/Desktop/~/Desktop/work/brain/fiftyone-brain/fiftyone/brain/internal/core/leaky_splits.py:380)         )
    [382](https://file+.vscode-resource.vscode-cdn.net/Users/jacobmarks/Desktop/~/Desktop/work/brain/fiftyone-brain/fiftyone/brain/internal/core/leaky_splits.py:382) return views

AttributeError: 'str' object has no attribute 'match_tags'
```


When running this code:

```py
dataset = foz.load_zoo_dataset("coco-2017", split="validation")
view = dataset.limit(10)
view.tag_samples("test")
view.untag_samples("validation")
fob.compute_leaky_splits(dataset, split_tags=["validation", "test"])
```

fiftyone/brain/internal/core/leaky_splits.py
```diff
@@ -0,0 +1,382 @@
...
+    results = brain_method.initialize(samples, brain_key)
+
+    if results._save_similarity_index:
+        results._similarity_method.register_run(
+            samples,
+            results.config.similarity_brain_key,
+        )
+
+    results.set_threshold(threshold)
+    leaks = results.leaks
```
jacobmarks, line 68: Getting this error:

```
File ~/Desktop/work/brain/fiftyone-brain/fiftyone/brain/similarity.py:856, in SimilarityIndex.get_model(self)
    [854](https://file+.vscode-resource.vscode-cdn.net/Users/jacobmarks/Desktop/~/Desktop/work/brain/fiftyone-brain/fiftyone/brain/similarity.py:854) model = self.config.model
    [855](https://file+.vscode-resource.vscode-cdn.net/Users/jacobmarks/Desktop/~/Desktop/work/brain/fiftyone-brain/fiftyone/brain/similarity.py:855) if model is None:
--> [856](https://file+.vscode-resource.vscode-cdn.net/Users/jacobmarks/Desktop/~/Desktop/work/brain/fiftyone-brain/fiftyone/brain/similarity.py:856)     raise ValueError("These results don't have a stored model")
    [858](https://file+.vscode-resource.vscode-cdn.net/Users/jacobmarks/Desktop/~/Desktop/work/brain/fiftyone-brain/fiftyone/brain/similarity.py:858) if etau.is_str(model):
    [859](https://file+.vscode-resource.vscode-cdn.net/Users/jacobmarks/Desktop/~/Desktop/work/brain/fiftyone-brain/fiftyone/brain/similarity.py:859)     model_kwargs = self.config.model_kwargs or {}

ValueError: These results don't have a stored model
```

When running this:

```py
dataset = foz.load_zoo_dataset("coco-2017", split="validation")
view1 = dataset.limit(1000)
view2 = dataset.skip(1000)

fob.compute_leaky_splits(dataset, split_views={"train":view1, "test":view2})
```

jacobmarks: Same thing happens when I add `similarity_brain_key="sim"`, and whether or not I include a `brain_key` arg

fiftyone/brain/internal/core/leaky_splits.py
jacobsela: Removed this check and computed embeddings on the fly if the index sizes don't line up. This leaves us open to the edge case where the previous run was on a different view with the same size we wouldn't know. I'm hesitant to replace the check from view size to dataset size because that could be a lot of unwanted compute. What do you guys think?

fiftyone/brain/internal/core/leaky_splits.py
jacobsela: changed

jacobmarks (changes requested):

fiftyone/brain/internal/core/leaky_splits.py
jacobsela: I never considered the option of someone not passing a model at all. I'll add a default.

jacobsela: Fixed how default values are handled, should work now. For what it's worth I highly recommend computing at least the embeddings ahead of time so it's batched and much much faster.

fiftyone/brain/internal/core/leaky_splits.py
```diff
@@ -0,0 +1,383 @@
...
+        super().__init__(samples, config, brain_key)
+
+        # process arguments to work only with views
+        self.split_views = _to_views(
+            samples,
+            self.config.split_views,
+            self.config.split_field,
+            self.config.split_tags,
+        )
+        self.id2split = self._id2splitConstructor()
```
jacobmarks, line 132: I'd make this internal

fiftyone/brain/__init__.py
```diff
@@ -703,3 +703,99 @@ def compute_exact_duplicates(
...
+
+def compute_leaky_splits(
+    samples,
+    brain_key=None,
+    split_views=None,
+    split_field=None,
+    split_tags=None,
+    threshold=0.2,
+    similarity_brain_key=None,
+    embeddings_field=None,
```
jacobmarks, line 716: Currently, this code results in all sample embeddings being computed:

```py
view1 = dataset.limit(500)
view2 = dataset.skip(500).limit(500)

fob.compute_leaky_splits(dataset, split_views={"train":view1, "test":view2}, brain_key="leak_key2", embeddings_field="clip_embeddings")
```

When in practice only 1000 need embeddings. I think this is fine, just calling it out

fiftyone/brain/internal/core/leaky_splits.py
```diff
@@ -0,0 +1,383 @@
...
+        duplicates = self._similarity_index.duplicates_view()
+
+        # filter duplicates to just those with neighbors in different splits
+        to_keep = []
+        for (
+            sample_id,
+            neighbors,
+        ) in self._similarity_index.neighbors_map.items():
+            keep_sample = False
+            sample_split = self.id2split[sample_id]
```
jacobmarks, line 231:
```py

view1 = dataset.limit(500)
view2 = dataset.skip(500).limit(500)

fob.compute_leaky_splits(dataset, split_views={"train":view1, "test":view2}, brain_key="leak_key2", embeddings_field="clip_embeddings")
```

Results in the following error:

```
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
Cell In[23], [line 4](vscode-notebook-cell:?execution_count=23&line=4)
      [1](vscode-notebook-cell:?execution_count=23&line=1) view1 = dataset.limit(500)
      [2](vscode-notebook-cell:?execution_count=23&line=2) view2 = dataset.skip(500).limit(500)
----> [4](vscode-notebook-cell:?execution_count=23&line=4) fob.compute_leaky_splits(dataset, split_views={"train":view1, "test":view2}, brain_key="leak_key2", embeddings_field="clip_embeddings")

File ~/Desktop/work/brain/fiftyone-brain/fiftyone/brain/__init__.py:787, in compute_leaky_splits(samples, brain_key, split_views, split_field, split_tags, threshold, similarity_brain_key, embeddings_field, model, model_kwargs, similarity_backend, similarity_config_dict, **kwargs)
    [723](https://file+.vscode-resource.vscode-cdn.net/Users/jacobmarks/Desktop/~/Desktop/work/brain/fiftyone-brain/fiftyone/brain/__init__.py:723) """Uses a similarity index or creates one on the spot to find leaks.
    [724](https://file+.vscode-resource.vscode-cdn.net/Users/jacobmarks/Desktop/~/Desktop/work/brain/fiftyone-brain/fiftyone/brain/__init__.py:724) 
    [725](https://file+.vscode-resource.vscode-cdn.net/Users/jacobmarks/Desktop/~/Desktop/work/brain/fiftyone-brain/fiftyone/brain/__init__.py:725) Calling this method only creates the index. You can then call the methods
   (...)
    [782](https://file+.vscode-resource.vscode-cdn.net/Users/jacobmarks/Desktop/~/Desktop/work/brain/fiftyone-brain/fiftyone/brain/__init__.py:782)     a :class:`fiftyone.core.view.DatasetView`
    [783](https://file+.vscode-resource.vscode-cdn.net/Users/jacobmarks/Desktop/~/Desktop/work/brain/fiftyone-brain/fiftyone/brain/__init__.py:783) """
    [785](https://file+.vscode-resource.vscode-cdn.net/Users/jacobmarks/Desktop/~/Desktop/work/brain/fiftyone-brain/fiftyone/brain/__init__.py:785) from fiftyone.brain.internal.core.leaky_splits import compute_leaky_splits
--> [787](https://file+.vscode-resource.vscode-cdn.net/Users/jacobmarks/Desktop/~/Desktop/work/brain/fiftyone-brain/fiftyone/brain/__init__.py:787) return compute_leaky_splits(
    [788](https://file+.vscode-resource.vscode-cdn.net/Users/jacobmarks/Desktop/~/Desktop/work/brain/fiftyone-brain/fiftyone/brain/__init__.py:788)     samples,
    [789](https://file+.vscode-resource.vscode-cdn.net/Users/jacobmarks/Desktop/~/Desktop/work/brain/fiftyone-brain/fiftyone/brain/__init__.py:789)     brain_key=brain_key,
    [790](https://file+.vscode-resource.vscode-cdn.net/Users/jacobmarks/Desktop/~/Desktop/work/brain/fiftyone-brain/fiftyone/brain/__init__.py:790)     split_views=split_views,
    [791](https://file+.vscode-resource.vscode-cdn.net/Users/jacobmarks/Desktop/~/Desktop/work/brain/fiftyone-brain/fiftyone/brain/__init__.py:791)     split_field=split_field,
    [792](https://file+.vscode-resource.vscode-cdn.net/Users/jacobmarks/Desktop/~/Desktop/work/brain/fiftyone-brain/fiftyone/brain/__init__.py:792)     split_tags=split_tags,
    [793](https://file+.vscode-resource.vscode-cdn.net/Users/jacobmarks/Desktop/~/Desktop/work/brain/fiftyone-brain/fiftyone/brain/__init__.py:793)     threshold=threshold,
    [794](https://file+.vscode-resource.vscode-cdn.net/Users/jacobmarks/Desktop/~/Desktop/work/brain/fiftyone-brain/fiftyone/brain/__init__.py:794)     similarity_brain_key=similarity_brain_key,
    [795](https://file+.vscode-resource.vscode-cdn.net/Users/jacobmarks/Desktop/~/Desktop/work/brain/fiftyone-brain/fiftyone/brain/__init__.py:795)     embeddings_field=embeddings_field,
    [796](https://file+.vscode-resource.vscode-cdn.net/Users/jacobmarks/Desktop/~/Desktop/work/brain/fiftyone-brain/fiftyone/brain/__init__.py:796)     model=model,
    [797](https://file+.vscode-resource.vscode-cdn.net/Users/jacobmarks/Desktop/~/Desktop/work/brain/fiftyone-brain/fiftyone/brain/__init__.py:797)     model_kwargs=model_kwargs,
    [798](https://file+.vscode-resource.vscode-cdn.net/Users/jacobmarks/Desktop/~/Desktop/work/brain/fiftyone-brain/fiftyone/brain/__init__.py:798)     similarity_backend=similarity_backend,
    [799](https://file+.vscode-resource.vscode-cdn.net/Users/jacobmarks/Desktop/~/Desktop/work/brain/fiftyone-brain/fiftyone/brain/__init__.py:799)     similarity_config_dict=similarity_config_dict,
    [800](https://file+.vscode-resource.vscode-cdn.net/Users/jacobmarks/Desktop/~/Desktop/work/brain/fiftyone-brain/fiftyone/brain/__init__.py:800)     **kwargs,
    [801](https://file+.vscode-resource.vscode-cdn.net/Users/jacobmarks/Desktop/~/Desktop/work/brain/fiftyone-brain/fiftyone/brain/__init__.py:801) )

File ~/Desktop/work/brain/fiftyone-brain/fiftyone/brain/internal/core/leaky_splits.py:67, in compute_leaky_splits(samples, brain_key, split_views, split_field, split_tags, threshold, similarity_brain_key, embeddings_field, model, model_kwargs, similarity_backend, similarity_config_dict, **kwargs)
     [61](https://file+.vscode-resource.vscode-cdn.net/Users/jacobmarks/Desktop/~/Desktop/work/brain/fiftyone-brain/fiftyone/brain/internal/core/leaky_splits.py:61) if results._save_similarity_index:
     [62](https://file+.vscode-resource.vscode-cdn.net/Users/jacobmarks/Desktop/~/Desktop/work/brain/fiftyone-brain/fiftyone/brain/internal/core/leaky_splits.py:62)     results._similarity_method.register_run(
     [63](https://file+.vscode-resource.vscode-cdn.net/Users/jacobmarks/Desktop/~/Desktop/work/brain/fiftyone-brain/fiftyone/brain/internal/core/leaky_splits.py:63)         samples,
     [64](https://file+.vscode-resource.vscode-cdn.net/Users/jacobmarks/Desktop/~/Desktop/work/brain/fiftyone-brain/fiftyone/brain/internal/core/leaky_splits.py:64)         results.config.similarity_brain_key,
     [65](https://file+.vscode-resource.vscode-cdn.net/Users/jacobmarks/Desktop/~/Desktop/work/brain/fiftyone-brain/fiftyone/brain/internal/core/leaky_splits.py:65)     )
---> [67](https://file+.vscode-resource.vscode-cdn.net/Users/jacobmarks/Desktop/~/Desktop/work/brain/fiftyone-brain/fiftyone/brain/internal/core/leaky_splits.py:67) results.set_threshold(threshold)
     [68](https://file+.vscode-resource.vscode-cdn.net/Users/jacobmarks/Desktop/~/Desktop/work/brain/fiftyone-brain/fiftyone/brain/internal/core/leaky_splits.py:68) leaks = results.leaks
     [70](https://file+.vscode-resource.vscode-cdn.net/Users/jacobmarks/Desktop/~/Desktop/work/brain/fiftyone-brain/fiftyone/brain/internal/core/leaky_splits.py:70) if brain_key is not None:

File ~/Desktop/work/brain/fiftyone-brain/fiftyone/brain/internal/core/leaky_splits.py:231, in leaks(self)
    [226](https://file+.vscode-resource.vscode-cdn.net/Users/jacobmarks/Desktop/~/Desktop/work/brain/fiftyone-brain/fiftyone/brain/internal/core/leaky_splits.py:226) to_keep = []
    [227](https://file+.vscode-resource.vscode-cdn.net/Users/jacobmarks/Desktop/~/Desktop/work/brain/fiftyone-brain/fiftyone/brain/internal/core/leaky_splits.py:227) for (
    [228](https://file+.vscode-resource.vscode-cdn.net/Users/jacobmarks/Desktop/~/Desktop/work/brain/fiftyone-brain/fiftyone/brain/internal/core/leaky_splits.py:228)     sample_id,
    [229](https://file+.vscode-resource.vscode-cdn.net/Users/jacobmarks/Desktop/~/Desktop/work/brain/fiftyone-brain/fiftyone/brain/internal/core/leaky_splits.py:229)     neighbors,
    [230](https://file+.vscode-resource.vscode-cdn.net/Users/jacobmarks/Desktop/~/Desktop/work/brain/fiftyone-brain/fiftyone/brain/internal/core/leaky_splits.py:230) ) in self._similarity_index.neighbors_map.items():
--> [231](https://file+.vscode-resource.vscode-cdn.net/Users/jacobmarks/Desktop/~/Desktop/work/brain/fiftyone-brain/fiftyone/brain/internal/core/leaky_splits.py:231)     keep_sample = False
    [232](https://file+.vscode-resource.vscode-cdn.net/Users/jacobmarks/Desktop/~/Desktop/work/brain/fiftyone-brain/fiftyone/brain/internal/core/leaky_splits.py:232)     sample_split = self.id2split[sample_id]
    [233](https://file+.vscode-resource.vscode-cdn.net/Users/jacobmarks/Desktop/~/Desktop/work/brain/fiftyone-brain/fiftyone/brain/internal/core/leaky_splits.py:233)     leaks = []

KeyError: '673f6702a7ac3f16db0b8c32'
```

After making the following progress:
```plaintext
Computing embeddings...
 100% |███████████████| 5000/5000 [5.1m elapsed, 0s remaining, 17.4 samples/s]      
Computing duplicate samples...
Generating index for 5000 embeddings...
Index complete
Generating neighbors graph for 4765 embeddings...
Index complete
Duplicates computation complete
```

jacobmarks (changes requested):

fiftyone/brain/internal/core/leaky_splits.py
jacobsela: This happens because self.samples has more samples than the union of the splits. The id2split dict is constructed on the splits, but the index is constructed on all of the self.samples. As we discussed above, the check to see if the union of the splits and self.samples is the same is not scalable. I can add another heuristic check (i.e. size), but I don't really like those. Honestly I'm fine wrapping id2split in a try except and if it fails throwing an error with something along the lines of "KeyError ... This usually happens when the samples, split, and similarity index don't line up. Please make sure the samples are the union of the splits and that if a similarity index is provided it was applied on a superset of samples". What do you guys think @jacobmarks @brimoor

fiftyone/brain/__init__.py
jacobsela: See comment below.

fiftyone/brain/internal/core/leaky_splits.py
jacobsela: changed

fiftyone/brain/internal/core/leaky_splits.py
```diff
@@ -0,0 +1,421 @@
...
+        **kwargs,
+    )
+    brain_method = config.build()
+    brain_method.ensure_requirements()
+
+    if brain_key is not None:
+        # Don't allow overwriting an existing run with same key, since we
+        # need the existing run in order to perform workflows like
+        # automatically cleaning up the backend's index
+        brain_method.register_run(samples, brain_key, overwrite=False)
```
jacobmarks, line 57: You should check to make sure this doesn't break anything, but by adding `cleanup=False` in `register_run()` and commenting out the `cleanup()` method in `LeakySplits` below, I was able to get rid of the error with `Limit` not being registered

fiftyone/brain/internal/core/leaky_splits.py
```diff
@@ -0,0 +1,421 @@
...
+        other_tags = [t for t in tags if not t == tag]
+        if len(view.match_tags(other_tags)) > 0:
+            raise ValueError(
+                f"One or more samples have more than one of the tags provided! Every sample should have at most one of the tags provided."
+            )
+
+    return views
+
+
+def _throw_index_is_bigger_warning(sample_id):
```
jacobmarks, line 417: I like this idea, but it gets very annoying when this happens 300 times in a row

jacobsela: I now make it very explicit in the docs that the similarity's samples should be equal to the samples passed to leaky splits which should equal the union of the splits. I think if the user chooses to ignore that they should either deal with the annoyance or suppress warnings. What's the alternative for people that make the mistake unwittingly and need an error?

fiftyone/brain/internal/core/leaky_splits.py
jacobsela: This works for registration and shouldn't break anything as far as I can tell but it doesn't fix the core issue when deleting a run.

fiftyone/brain/internal/core/leaky_splits.py
```diff
@@ -0,0 +1,421 @@
...
+    def leaks_for_sample(self, sample):
+        """Return view with all leaks related to a certain sample.
+
+        Args:
+            sample: sample object or sample id
+
+        """
+        # compute leaks if it hasn't happend yet
+        if not self._leak_threshold == self._last_computed_threshold:
+            leaks = self.leaks
```
jacobmarks, line 301: Is this variable used? Looks like it isn't

fiftyone/brain/internal/core/leaky_splits.py
```diff
@@ -0,0 +1,421 @@
...
+    views = {}
+    for tag in tags:
+        view = samples.match_tags([tag])
+        if len(view) < 1:  # no samples in tag
+            raise ValueError(
+                f"One of the tags provided, '{tag}', has no samples. Make sure every tag has at least one sample."
+            )
+        views[tag] = view
+
+    for i, (tag, view) in enumerate(views.items()):
```
jacobmarks, line 407: index `i` isn't used

jacobmarks (commented):
This is looking very solid! I think it's almost at the finish line. 

I've tested the following:

- using `split_field`, `split_tags`, and `split_views` to specify splits, with or without all samples included
- using existing embeddings or not; using existing similarity index or not
- edge cases where just one split is given, and where every sample is assigned to its own split. In the former, an error is thrown as is appropriate. In the latter case, leaky splits works (although is slow, as is to be expected)

Currently testing on a larger dataset to make sure nothing breaks, but getting very close. 

Minor nit @jacobsela :make sure to lint all code before merging the PR


@mwoodson1 can you take a look at this PR as well? there are a lot of moving pieces so more people trying to break things now will be better in the end

fiftyone/brain/internal/core/leaky_splits.py
jacobsela: it can be an underscore instead. Really I'm just trying to invoke the self.leaks property so that the computation occurs.

jacobsela: made the change

jacobmarks:
Noting that
- everything works as expected at the scale of 50k samples.
- `threshold` control works as expected
- `tag_leaks()` works as expected
- `index.leaks` works as expected

I was a bit surprised by the behavior of `no_leaks_view()`, that it requires a view as input. Need to think on this

fiftyone/brain/internal/core/leaky_splits.py
jacobsela: removed

jacobsela:
@jacobmarks This was also something I thought about. The issue with having no arguments is that I need to know where the user wants to keep images. Just removing all leaks from every split wastes data. I can maybe make it so that the argument is the split to keep/remove leaks from. What do you think?

fiftyone/brain/internal/core/leaky_splits.py
jacobmarks: Really? It did for me, at least when running `dataset.delete_brain_runs()`

fiftyone/brain/internal/core/leaky_splits.py
```diff
@@ -0,0 +1,421 @@
...
+            self._similarity_index = self._similarity_method.initialize(
+                samples, self.config.similarity_brain_key
+            )
+
+    def set_threshold(self, threshold):
+        """Set threshold for leak computation"""
+        self._leak_threshold = threshold
+
+    @property
+    def leaks(self):
```
jacobmarks, line 226: For consistency with `no_leaks_view` here and `duplicates_view` in similarity API, please change this to `leaks_view`

fiftyone/brain/internal/core/leaky_splits.py
```diff
@@ -0,0 +1,421 @@
...
+                    ]
+                    neighbors_ids.append(unique_id)
+                    break
+
+        return self.samples.select([sample_id] + neighbors_ids)
+
+    def no_leaks_view(self, view):
+        return view.exclude(self.leaks.values("id"))
+
+    def tag_leaks(self, tag="leak"):
```
jacobmarks, line 332: You will need to change `leaks` to `leaks_view` here as well. Also, instead of iterating over all samples and saving, please use [`tag_samples` ()](https://docs.voxel51.com/api/fiftyone.core.collections.html#fiftyone.core.collections.SampleCollection.tag_samples)

fiftyone/brain/internal/core/leaky_splits.py
```diff
@@ -0,0 +1,421 @@
...
+            sample_ids = split_view.values("id")
+            id2split.update({sid: split_name for sid in sample_ids})
+
+        return id2split
+
+
+def _to_views(samples, split_views=None, split_field=None, split_tags=None):
+    """Helper function so that we can always work with views"""
+
+    arithmetic_true = lambda x: int(x is not None)
```
jacobmarks, line 355: This is not descriptive enough.

fiftyone/brain/internal/core/leaky_splits.py
```diff
@@ -0,0 +1,421 @@
...
+"""
+
+from copy import copy
+import warnings
+
+from fiftyone import ViewField as F
+import fiftyone.core.brain as fob
+import fiftyone.brain.similarity as sim
+import fiftyone.core.utils as fou
+import fiftyone.core.validation as fov
```
jacobmarks, line 12: Alphabetize

jacobmarks (approved):
Overall LGTM. Please make the changes requested before merging into main:

- change `leaks` to `leaks_view`
- implementation of tagging

Also make sure to add documentation and unit testing to the main FO repo. Please request me as a reviewer on those PRs

jacobsela:
Link to docs PR: https://github.com/voxel51/fiftyone/pull/5189

PR #204, opened 2024-11-03
Adding similarity_index parameter to all relevant methods

brimoor:
## Change log

- Adds the `similarity_index` parameter introduced in #201 to `compute_visualization()` and `compute_representativeness()`.
- Tweaks the implementation of `compute_uniqueness(similarity_index=)` in #201 to support the case where the index doesn't contain all samples in the dataset. A test case is added to verify this

## Example usage

```py
import fiftyone as fo
import fiftyone.zoo as foz
import fiftyone.brain as fob

dataset.delete_sample_field("uniqueness")

view = dataset.limit(100)
similarity_index = fob.compute_similarity(view)
# similarity_index = fob.compute_similarity(view, backend="qdrant")

results = fob.compute_visualization(dataset, similarity_index=similarity_index)
assert len(results.points) == len(view)

fob.compute_uniqueness(dataset, similarity_index=similarity_index)
assert len(dataset.exists("uniqueness")) == len(view)

fob.compute_representativeness(dataset, similarity_index=similarity_index)
assert len(dataset.exists("representativeness")) == len(view)
```

### DIFF
--- a/fiftyone/brain/__init__.py
+++ b/fiftyone/brain/__init__.py
@@ -176,6 +176,7 @@ def compute_uniqueness(
     uniqueness_field="uniqueness",
     roi_field=None,
     embeddings=None,
+    similarity_index=None,
     model=None,
     model_kwargs=None,
     force_square=False,
@@ -184,8 +185,6 @@ def compute_uniqueness(
     num_workers=None,
     skip_failures=True,
     progress=None,
-    similarity_backend=None,
-    similarity_index=None,
 ):
     """Adds a uniqueness field to each sample scoring how unique it is with
     respect to the rest of the samples.
@@ -193,8 +192,8 @@ def compute_uniqueness(
     This function only uses the pixel data and can therefore process labeled or
     unlabeled samples.
 
-    If no ``embeddings`` or ``model`` is provided, a default model is used to
-    generate embeddings.
+    If no ``embeddings``, ``similarity_index``, or ``model`` is provided, a
+    default model is used to generate embeddings.
 
     .. note::
 
@@ -223,6 +222,9 @@ def compute_uniqueness(
             when working with patch embeddings, you can provide either the
             fully-qualified path to the patch embeddings or just the name of
             the label attribute in ``roi_field``
+        similarity_index (None): a
+            :class:`fiftyone.brain.similarity.SimilarityIndex` or the brain key
+            of a similarity index to use to load pre-computed embeddings
         model (None): a :class:`fiftyone.core.models.Model` or the name of a
             model from the
             `FiftyOne Model Zoo <https://docs.voxel51.com/user_guide/model_zoo/models.html>`_
@@ -250,11 +252,6 @@ def compute_uniqueness(
         progress (None): whether to render a progress bar (True/False), use the
             default value ``fiftyone.config.show_progress_bars`` (None), or a
             progress callback function to invoke instead
-        similarity_backend (None): the similarity backend to use. The supported values are
-            ``fiftyone.brain.brain_config.similarity_backends.keys()`` and the
-            default is ``fiftyone.brain.brain_config.default_similarity_backend``
-        similarity_index (None): the similarity index to use. This can be set to
-            ``fiftyone.brain.similarity.SimilarityIndex`` instance or a brain_key string to load the similarity index via samples.load_brain_results
     """
     import fiftyone.brain.internal.core.uniqueness as fbu
 
@@ -263,6 +260,7 @@ def compute_uniqueness(
         uniqueness_field,
         roi_field,
         embeddings,
+        similarity_index,
         model,
         model_kwargs,
         force_square,
@@ -271,8 +269,6 @@ def compute_uniqueness(
         num_workers,
         skip_failures,
         progress,
-        similarity_backend,
-        similarity_index,
     )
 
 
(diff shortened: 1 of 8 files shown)


fiftyone/brain/__init__.py
```diff
@@ -250,11 +252,6 @@ def compute_uniqueness(
         progress (None): whether to render a progress bar (True/False), use the
             default value ``fiftyone.config.show_progress_bars`` (None), or a
             progress callback function to invoke instead
-        similarity_backend (None): the similarity backend to use. The supported values are
```
brimoor, line 253: After sitting with this for a bit, I think it is cleaner to expect the user to create the `SimilarityIndex` offline first and pass it in via `similarity_index` if they want control over what index is used to compute uniqueness. One key reason is my comment below about how the embeddings unfortunately need to be downloaded anyway to figure out what samples exist in the index and can therefore have their uniqueness computed.

fiftyone/brain/internal/core/uniqueness.py
```diff
@@ -66,83 +63,73 @@ def compute_uniqueness(
...
+        if batch_size is None:
+            batch_size = _DEFAULT_BATCH_SIZE
+
+    if roi_field is not None:
+        # @todo experiment with mean(), max(), abs().max(), etc
+        agg_fcn = lambda e: np.mean(e, axis=0)
+    else:
+        agg_fcn = None
+
+    embeddings, sample_ids, _ = fbu.get_embeddings(
```
brimoor, line 104: Unfortunately, `get_embeddings()` needs to be called in all cases, even when a pre-existing `similarity_index` is provided, because we need to query the index to see if a given sample ID actually exists in the vector index. It may not if the index was generated for a subset of the dataset. A typical case for this would be if new samples have been added to the dataset.

fiftyone/brain/internal/core/uniqueness.py
```diff
@@ -166,20 +155,28 @@ def compute_uniqueness(
...
-    dists = np.array(dists_list)
+    num_embeddings = len(embeddings)
+    if num_embeddings <= K:
+        return [1] * num_embeddings
+
+    if similarity_index.config.method == "sklearn":
+        _, dists = similarity_index._kneighbors(k=K + 1, return_dists=True)
+    else:
+        dists = []
+        with fou.ProgressBar(total=num_embeddings, progress=progress) as pb:
```
brimoor, line 171: This lays the groundwork for batching. The `_kneighbors()` interface already allows it, but none of the backends currently do anything smarter than sequential queries. However, if/when backends do support batched queries, this will automatically leverage it.

mwoodson1 (approved):
Good catch on the use of a potentially different subsets index. Everything lgtm

fiftyone/brain/internal/core/uniqueness.py
manushreegangwar: Good point. IIUC this implementation will raise a warning about the missing samples and continue with the uniqueness computation for the subset of samples. What if a significant number of sample ids are missing from the provided similarity index? Do you still want this behavior?

fiftyone/brain/__init__.py
manushreegangwar: I am in favor of this. Keeps the implementation simpler.

manushreegangwar (commented):
Few questions and comments. Looks good otherwise.

fiftyone/brain/internal/core/uniqueness.py
brimoor: Yeah that would currently require the person running the test to have one of the non-builtin similarity backends installed. I did manually verify that the code works by running the code in the PR description with `backend="qdrant"`.

fiftyone/brain/internal/core/uniqueness.py
brimoor: Yes I think its better to compute uniqueness for everything we can rather than raise a fatal error.

fiftyone/brain/internal/core/uniqueness.py
```diff
@@ -166,20 +155,28 @@ def compute_uniqueness(
...
 
 
-def _compute_uniqueness(sample_ids, similarity_index, n_neighbors=3):
-    num_samples = len(sample_ids)
-    if num_samples <= n_neighbors:
-        return [1] * num_samples
+def _compute_uniqueness(
+    embeddings, similarity_index, batch_size=10, progress=None
+):
+    K = 3
```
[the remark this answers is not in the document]
brimoor: `K` can't be independently changed. The hard-coded weights `[0.6, 0.3, 0.1]` implicitly assume `K=3`.

fiftyone/brain/internal/core/uniqueness.py
```diff
@@ -66,83 +63,73 @@ def compute_uniqueness(
...
+        agg_fcn=agg_fcn,
+        batch_size=batch_size,
+        num_workers=num_workers,
+        skip_failures=skip_failures,
+        progress=progress,
+    )
+
+    if similarity_index is None:
+        similarity_index = fb.compute_similarity(
+            samples, backend="sklearn", embeddings=False
```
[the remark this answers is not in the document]
brimoor: Not necessarily! Users can configure a different default backend as described here: 
https://docs.voxel51.com/brain.html#changing-your-similarity-backend

But in this context, we always want to use `sklearn`

PR #212, opened 2024-11-29
Leaky splits updates (option 1)

brimoor:
## Change log

- removed `brain_key` argument from `compute_leaky_splits()` as there is no functional value in saving a leaky splits index on a dataset (it doesn't directly persist anything, it just uses a similarity index)
- use `embeddings` and `similarity_index` argument naming/functionality for consistency with other methods like `compute_uniqueness()` and `compute_representativeness()`
- only return `index` for from `compute_leaky_splits()` for consistency with other brain methods
- added `thresh` property to index for consistency with `DuplicatesMixin.thresh`
- added `leak_ids` property to index for consistency with `DuplicatesMixin.duplicate_ids`
- consolidated `split_tags|split_fields|split_views` into a single `splits` argument, for consistency with existing conventions of having arguments tied to function (splits) rather than form (how the splits are specified)
- optimized the implementation of `leaks_view()` and `no_leaks_view()`
- added (back) warnings when the input splits contain overlaps or the similarity index doesn't contain samples from the splits

## Example usage

```py
import fiftyone as fo
import fiftyone.brain as fob
import fiftyone.zoo as foz
import fiftyone.utils.random as four

# Load some COCO data
dataset = foz.load_zoo_dataset(
    "coco-2017",
    split="validation",
    max_samples=1000,
    persistent=True,
)

# Set up splits via tags
dataset.untag_samples(dataset.distinct("tags"))
four.random_split(dataset, {"train": 0.7, "test": 0.3})

# Compute embeddings
model = foz.load_zoo_model("mobilenet-v2-imagenet-torch")
dataset.compute_embeddings(model, embeddings_field="embeddings")

# Find leaks
index = fob.compute_leaky_splits(dataset, splits=["train", "test"], embeddings="embeddings")

print(index.thresh)  # 0.2
print(index.leak_ids)  # ['XXXX', ...]

leaks_view = index.leaks_view()
no_leaks_view = index.no_leaks_view()
leaks_for_sample_view = index.leaks_for_sample(index.leak_ids[0])

session = fo.launch_app(leaks_view)
```

### DIFF
--- a/fiftyone/brain/__init__.py
+++ b/fiftyone/brain/__init__.py
@@ -712,103 +712,98 @@ def compute_exact_duplicates(
 
 def compute_leaky_splits(
     samples,
-    brain_key=None,
-    split_views=None,
-    split_field=None,
     split_tags=None,
+    split_field=None,
+    split_views=None,
     threshold=0.2,
-    similarity_brain_key=None,
-    embeddings_field=None,
+    embeddings=None,
+    similarity_index=None,
     model=None,
     model_kwargs=None,
-    similarity_backend=None,
-    similarity_config_dict=None,
-    **kwargs,
+    batch_size=None,
+    num_workers=None,
+    skip_failures=True,
+    progress=None,
 ):
-    """Uses a similarity index or creates one on the spot to find leaks.
+    """Computes potential leaks between splits of the given sample collection.
 
-    Calling this method only creates the index. You can then call the methods
-    exposed on the returned object to perform the following operations:
+    You must provide one of ``split_tags``, ``split_field``, or ``split_views``
+    to use this method. It is assumed that the specified splits fully partition
+    the input collection (disjoint and fully covering).
 
-    -   :meth:`leaks_view <fiftyone.brain.core.internal.leaky_splits.LeakySplitIndex.leaks>`:
-        Returns a view of all leaks in the dataset.
+    Calling this method only initializes the index. You can then call the
+    methods exposed on the returned object to perform the following operations:
 
-    -   :meth:`no_leaks_view <fiftyone.brain.core.internal.leaky_splits.LeakySplitIndex.no_leaks_view>`:
-        Returns a subset of the given view without any leaks.
+    -   :meth:`leaks_view() <fiftyone.brain.core.internal.leaky_splits.LeakySplitIndex.leaks_view>`:
+        Returns a view of all leaks in the input collection
 
-    -   :meth:`leaks_for_sample <fiftyone.brain.core.internal.leaky_splits.LeakySplitIndex.leaks_for_sample>`:
+    -   :meth:`no_leaks_view() <fiftyone.brain.core.internal.leaky_splits.LeakySplitIndex.no_leaks_view>`:
+        Returns the subset of the input collection without any leaks
+
+    -   :meth:`leaks_for_sample() <fiftyone.brain.core.internal.leaky_splits.LeakySplitIndex.leaks_for_sample>`:
         Returns a view with leaks corresponding to the given sample.
 
-    -   :meth:`tag_leaks <fiftyone.brain.core.internal.leaky_splits.LeakySplitIndex.tag_leaks>`:
+    -   :meth:`tag_leaks() <fiftyone.brain.core.internal.leaky_splits.LeakySplitIndex.tag_leaks>`:
         Tags leaks in the dataset as leaks.
 
-
     Args:
-        samples: a :class:`fiftyone.core.collections.SampleCollection` This should be a union of the
-            splits provided.
-        brain_key (None): a brain key under which to store the results of this
-            method. If no brain key is provided the results will not be saved.
-        split_views (None): a dict of :class:`fiftyone.core.view.DatasetView`
-            corresponding to different splits in the datset. Only one of
-            `split_views`, `split_field`, and `splits_tags` need to be used.
-        split_field (None): a string name of a field that holds the split of the sample.
-            Each unique value in the field will be treated as a split.
-            Only one of `split_views`, `split_field`, and `splits_tags` need to be used.
-        split_tags (None): a list of strings, tags corresponding to differents splits.
-            Only one of `split_views`, `split_field`, and `splits_tags` need to be used.
+        samples: a :class:`fiftyone.core.collections.SampleCollection`
+        split_tags (None): a list of tag strings
+        split_field (None): the name of a string field that encodes the split
+            memberships
+        split_views (None): a dict mapping split names to
+            :class:`fiftyone.core.view.DatasetView` instances
+        threshold (0.2): the similarity distance threshold to use when
+            detecting leaks. Values in ``[0.1, 0.25]`` work well for the
+            default setup
+        embeddings (None): if no ``model`` is provided, this argument specifies
+            pre-computed embeddings to use, which can be any of the following:
 
-        The splits should be disjoint and their union should be samples.
+            -   a ``num_samples x num_dims`` array of embeddings
+            -   if ``roi_field`` is specified,  a dict mapping sample IDs to
+                ``num_patches x num_dims`` arrays of patch embeddings
+            -   the name of a dataset field containing the embeddings to use
 
-        threshold (0.2): The threshold to run the algorithm with. Values between
-            0.1 - 0.25 tend to give good results.
-        similarity_brain_key (None): a brain key for the similarity index
-            If the brain key exists already, it will load up the similarity index corresponding to it
-            If the brain key does not exist already, a new similarity index will be created
-            and the results will be saved under this name
-
-            The similarity backend passed should have been computed on at least the argument samples.
-            This method may break if this condition is not met.
-
-        embeddings_field (None): field for embeddings to feed the index. This argument's
-            behavior depends on whether a ``model`` is provided, as described
-            below.
-            If no ``model`` is provided, this argument specifies the field of pre-computed
-            embeddings to use
-            If a ``model`` is provided, this argument specifies where to store
-            the model's embeddings
+            If a ``model`` is provided, this argument specifies the name of a
+            field in which to store the computed embeddings
+        similarity_index (None): a
+            :class:`fiftyone.brain.similarity.SimilarityIndex` or the brain key
+            of a similarity index to use to load pre-computed embeddings
         model (None): a :class:`fiftyone.core.models.Model` or the name of a
             model from the
-            `FiftyOne Model Zoo <https://docs.voxel51.com/user_guide/model_zoo/index.html>`_
-            to use, or that was already used, to generate embeddings. The model
-            must expose embeddings (``model.has_embeddings = True``)
+            `FiftyOne Model Zoo <https://docs.voxel51.com/user_guide/model_zoo/models.html>`_
+            to use to generate embeddings. The model must expose embeddings
+            (``model.has_embeddings = True``)
         model_kwargs (None): a dictionary of optional keyword arguments to pass
             to the model's ``Config`` when a model name is provided
-        similarity_backend: string, the similarity backend to use. The supported values are
-            ``fiftyone.brain.brain_config.similarity_backends.keys()`` and the
-            default is
-            ``fiftyone.brain.brain_config.default_similarity_backend``
-        similarity_config_dict: dict, used to build the similarity backend. Arguments take
-            precedence over the values in the dict (e.g. model)
+        batch_size (None): a batch size to use when computing embeddings. Only
+            applicable when a ``model`` is provided
+        num_workers (None): the number of workers to use when loading images.
+            Only applicable when a Torch-based model is being used to compute
+            embeddings
+        skip_failures (True): whether to gracefully continue without raising an
+            error if embeddings cannot be generated for a sample
+        progress (None): whether to render a progress bar (True/False), use the
+            default value ``fiftyone.config.show_progress_bars`` (None), or a
+            progress callback function to invoke instead
 
     Returns:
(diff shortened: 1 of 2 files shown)


jacobsela:
This is definitely a lot cleaner. I’ll give it a thorough look on Monday.

what do you think about having an EmbeddingsBrainMethodMixin or something similar that handles the boilerplate around embeddings and models? It seems strange to me to duplicate that code every time.

fiftyone/brain/__init__.py
```diff
@@ -712,103 +712,98 @@ def compute_exact_duplicates(
 
 def compute_leaky_splits(
     samples,
-    brain_key=None,
```
brimoor, line 715: Happy to chat more offline about the removal of `brain_key`, but my reasoning is that saving a `LeakySplitsIndex` as currently implemented doesn't persist anything of value. It doesn't store the embeddings directly or do any reusable computation that would compel users to call `compute_leaky_splits()` once and then reuse the same object in the future.

fiftyone/brain/__init__.py
```diff
@@ -712,103 +712,98 @@ def compute_exact_duplicates(
...
-    split_views=None,
-    split_field=None,
     split_tags=None,
+    split_field=None,
+    split_views=None,
     threshold=0.2,
-    similarity_brain_key=None,
```
brimoor, line 720: These changes:
- `embeddings_field` -> `embeddings`
- `similarity_brain_key` -> `similarity_index`

are for consistency with other methods like `compute_uniqueness()`.

The idea behind removing `similarity_backend` and `similarity_config_dict`, aside from also being consistent with other brain methods, is basically: if you want fine-grained control over the similarity index, just make one yourself and pass it in via `similarity_index`.

fiftyone/brain/internal/core/leaky_splits.py
```diff
@@ -1,108 +1,140 @@
...
-    results.set_threshold(threshold)
-    leaks = results.leaks_view()
-
-    if brain_key is not None:
-        brain_method.save_run_results(samples, brain_key, results)
-    if results._save_similarity_index:
-        results._similarity_method.save_run_results(
-            samples, similarity_brain_key, results._similarity_index
+    elif not isinstance(similarity_index, fbs.DuplicatesMixin):
+        raise ValueError(
```
brimoor, line 94: Important to explicitly validate this; wasn't being done previously

fiftyone/brain/internal/core/leaky_splits.py
```diff
@@ -1,108 +1,140 @@
...
-        # need the existing run in order to perform workflows like
-        # automatically cleaning up the backend's index
-        brain_method.register_run(samples, brain_key, overwrite=False)
-
-    results = brain_method.initialize(samples, brain_key)
-
-    if results._save_similarity_index:
-        results._similarity_method.register_run(
+    if similarity_index is None:
+        similarity_index = fb.compute_similarity(
```
brimoor, line 82: Cleaner to call `compute_similarity()` to generate a similarity index rather than directly interacting the underlying classes as was done in the previous implementation

fiftyone/brain/internal/core/leaky_splits.py
```diff
@@ -1,108 +1,140 @@
...
-        split_field,
-        split_tags,
-        similarity_brain_key,
-        embeddings_field,
-        model,
-        model_kwargs,
-        similarity_backend,
-        similarity_config_dict,
-        **kwargs,
+        split_field=split_field,
```
brimoor, line 70: There's honestly not any value in much of this boilerplate code here (using the `BrainMethodConfig`/`BrainMethod`/`BrainResults` interface) as currently implemented as leaky split indexes are not registered on the dataset currently.

I left it in so I could explain a few things about how to use the interface (see below)

fiftyone/brain/internal/core/leaky_splits.py
```diff
@@ -1,108 +1,140 @@
...
         embeddings_field=None,
+        similarity_index=None,
         model=None,
         model_kwargs=None,
-        similarity_backend=None,
-        similarity_config_dict=None,
         **kwargs,
     ):
-        self.split_views = split_views
+        if similarity_index is not None and not etau.is_str(similarity_index):
```
brimoor, line 125: All public properties of `BrainMethodConfig` classes must be JSON serializable, as this classes' public properties are automatically serialized and made available  via `load_brain_info().config` (**IF** the brain run is stored under a `brain_key`)

fiftyone/brain/internal/core/leaky_splits.py
```diff
@@ -115,312 +147,241 @@ def method(self):
...
-        self.index._similarity_method.cleanup(samples, brain_key)
-        super.cleanup(samples, brain_key)
+        return fields
 
 
 class LeakySplitsIndex(fob.BrainResults):
-    def __init__(self, samples, config, brain_key):
-        super().__init__(samples, config, brain_key)
+    def __init__(self, samples, config, similarity_index, split_views):
+        super().__init__(samples, config, None)
```
brimoor, line 165: If we wanted to add support for `compute_leaky_splits(..., brain_key=...)`, then this class would need to be refactored such that this would work:

```py
index = dataset.load_brain_results(brain_key)
index.find_leaks()
```

That would be a reasonable thing to do **IF** the user originally provided a `similarity_index`, as we could just use `config.similarity_index` to load it.

If the user didn't provide a similarity index, then they would need to have provided either an `embeddings="field_name"` or `model="model_name"` so that we could load/recompute the necessary embeddings. However, by convention, `load_brain_results()` should really not do any *novel* compute, it is assumed that any expensive compute was done when the method was originally called and, as per the `BrainResults` class naming, the **resulting** data should be stored such that it can be loaded **quickly** via `load_brain_results()` in a new session.

The TLDR here is that `compute_similarity()` is really doing all the heavy-lifting here. `compute_leaky_splits()` is really just a function that you call that uses a similarity index to do something useful.

fiftyone/brain/internal/core/leaky_splits.py
```diff
@@ -115,312 +147,241 @@ def method(self):
...
+        Returns:
+            a :class:`fiftyone.core.view.DatasetView`
+        """
+        if self._thresh is None:
+            raise ValueError("You must first call `find_leaks()`")
 
-        # cache to avoid recomputation
-        self._last_computed_threshold = self._leak_threshold
-        self._leaks = duplicates
+        return self.samples.select(self._leak_ids, ordered=True)
```
brimoor, line 243: The previous implementation used:

```py
duplicates_view().select(to_keep)
```

which is not ideal because `duplicates_view()` internally performs a potentially much larger `select(..., ordered=True)` operation across all duplicates, not just specifically leaks.

Directly doing `select(self._leak_ids, ordered=True)` gets right to the point.

fiftyone/brain/internal/core/leaky_splits.py
```diff
@@ -115,312 +147,241 @@ def method(self):
...
-    num_given = (
-        arithmetic_true(split_views)
-        + arithmetic_true(split_field)
-        + arithmetic_true(split_tags)
-    )
+        # Check for overlapping splits
+        split_names = list(split_ids.keys())
+        for idx, split1 in enumerate(split_names):
+            for split2 in split_names[idx + 1 :]:
+                overlap = split_ids[split1] & split_ids[split2]
```
brimoor, line 323: You might have had a check like this in the initial implementation but I cautioned against it because loading all IDs into memory via `values("id")` can potentially be expensive for large datasets.

I added it back for two reasons:
- The implementation *already* requires calling `values("id")` for all splits, so might as well do the validation
- I wasn't aware that the implementation requires `DuplicatesMixin`, which is only implemented by the sklearn backend, which isn't really designed for large datasets anyway

fiftyone/brain/internal/core/leaky_splits.py
```diff
@@ -115,312 +147,241 @@ def method(self):
...
+                        split1,
+                        split2,
+                    )
 
-    if num_given == 0:
-        raise ValueError(f"One of the split arguments must be given.")
-    if num_given > 1:
-        raise ValueError(f"Only one of the split arguments must be given.")
+        # Check for samples not in index
+        index_ids = self._similarity_index.sample_ids
```
brimoor, line 337: The `SimilarityIndex` interface provides a `sample_ids` property:
https://github.com/voxel51/fiftyone-brain/blob/7b0259ce1bc1070050de998679ded723938bbf4f/fiftyone/brain/similarity.py#L339-L342

which is only implemented by backends for which getting the full set of IDs in the index is something that can be done in a reasonable amount of time. So, this implementation will perform the ID verification that you wanted **IF** the similarity backend supports it.

Now, in practice, the only backend that currently supports `SimiliarityIndex.sample_ids` is the sklearn backend, which not coincidentally is the only one that implementation `DuplicatesMixin`, so in practice this ID verification will now always happen.

fiftyone/brain/__init__.py
```diff
@@ -712,103 +712,98 @@ def compute_exact_duplicates(
...
+        skip_failures (True): whether to gracefully continue without raising an
+            error if embeddings cannot be generated for a sample
+        progress (None): whether to render a progress bar (True/False), use the
+            default value ``fiftyone.config.show_progress_bars`` (None), or a
+            progress callback function to invoke instead
 
     Returns:
-        a :class:`fiftyone.brain.internal.core.leaky_splits.LeakySplitsIndex`,
-        a :class:`fiftyone.core.view.DatasetView`
+        a :class:`fiftyone.brain.internal.core.leaky_splits.LeakySplitsIndex`
```
brimoor, line 791: This is a breaking change to the interface: users must now run:

```py
index = fob.compute_leaky_splits(dataset, ...)
leaks = index.leaks_view()
```

rather than

```py
index, leaks = fob.compute_leaky_splits(dataset, ...)
```

Reasoning here is for consistency with other brain methods, and also because there's no guarantee I want `leaks_view`: maybe I just want to access `index.leak_ids` or `index.no_leaks_view()`, for example.

I updated the Brain docs in https://github.com/voxel51/fiftyone/pull/5188 to account for this change, but I assume the blog post will need tweaked as well.

fiftyone/brain/internal/core/leaky_splits.py
```diff
@@ -1,108 +1,140 @@
...
+    num_workers=None,
+    skip_failures=True,
+    progress=None,
 ):
     """See ``fiftyone/brain/__init__.py``."""
 
-    fov.validate_collection(samples)
+    fov.validate_image_collection(samples)
+
+    if etau.is_str(embeddings):
```
brimoor, line 47: > what do you think about having an EmbeddingsBrainMethodMixin or something similar that handles the boilerplate around embeddings and models? It seems strange to me to duplicate that code every time.

Taking this conclusion to the logical extreme, the way to implement leaky splits with the absolute minimum amount of code would definitely have been to add it to the `DuplicatesMixin` interface, say like this:

```py
index = fob.compute_similarity(dataset, ...)

index.find_leaks(splits=["train", "test"], threshold=0.2)
index.find_leaks(splits="split_field", threshold=0.2)
index.find_leaks(splits={"train": train_view, "test": test_view", threshold=0.2)

print(index.leak_ids)  # ['XXXX', ...]

leaks = index.leaks_view()
no_leaks_view = index.no_leaks_view()
leaks_for_sample = index.leaks_for_sample(index.leak_ids[0])
```

After all, leaky splits is really just an add-on on top of the duplicates stuff, and that's exactly how those were implemented:

```py
index.find_duplicates()
index.duplicate_ids
index.duplicates_view()

index.find_unique()
index.unique_ids
index.unique_view()
```

Now having said that, I do see value in having a top-level `compute_leaky_splits()` method, as that's more of a use-case-centric naming convention (methods based on what you want to do), rather than an implementation-centric naming convention (methods based on how they work).

So, I'm onboard with `compute_leaky_splits()` and in fact think it could make sense in a follow-up PR to add a `compute_near_duplicates()` method that basically wraps `DuplicatesMixin.find_duplicates()`.

So, to your question: the only reason for the boilerplate code here is to parse out the JSON serializable config parameters to pass to `BrainMethod Config`. **BUT**, as discussed elsewhere on this PR, there's really not any need to do all that as currently implemented because there's no `brain_key`.

I'll make a follow-up PR that prunes the unnecessary boilerplate to see what we think about it.

brimoor: Follow-up PR is here: https://github.com/voxel51/fiftyone-brain/pull/213

fiftyone/brain/internal/core/leaky_splits.py
```diff
@@ -1,108 +1,134 @@
...
 import fiftyone.core.validation as fov
+import fiftyone.zoo as foz
+from fiftyone import ViewField as F
+
+import fiftyone.brain as fb
+import fiftyone.brain.similarity as fbs
+import fiftyone.brain.internal.core.utils as fbu
+
 
-fbu = fou.lazy_import("fiftyone.brain.internal.core.utils")
```
jacobsela, line 14: Why is this done as a lazy import in similarity?

jacobsela (approved):
I like the changes. This is cleaner.

jacobsela:
@brimoor I haven't had a chance to thoroughly test yet. If this is necessary please let me know and I'll prioritize it now.

fiftyone/brain/internal/core/leaky_splits.py
brimoor: The two main reasons `lazy_import()` is used are to avoid circular import errors and to avoid importing packages that aren't installed by default.

Can't remember why it was used in `fiftyone.brain.similarity` 🤷

PR #218, opened 2024-12-05
Upgrade MongoDB Atlas vector search integration to use `vectorSearch` type

brimoor:
## Change log

- upgrades the MongoDB Atlas vector search integration to use the generally available `vectorSearch` type rather than the beta `knnVector` type 
- fix bug with similarity queries on grouped datasets

## Example usage

```shell
export FIFTYONE_DATABASE_NAME=fiftyone
export FIFTYONE_DATABASE_URI='mongodb+srv://$USERNAME:$PASSWORD@$NAME.mongodb.net/?retryWrites=true&w=majority'
```

```py
import fiftyone as fo
import fiftyone.zoo as foz
import fiftyone.brain as fob

dataset = foz.load_zoo_dataset("quickstart-groups")

view = dataset.select_group_slices("left")
index = fob.compute_similarity(
    view,
    model="clip-vit-base32-torch",
    brain_key="img_sim",
    backend="mongodb",
    embeddings="embeddings",
)

assert index.ready

print(index.total_index_size)  # 200

view = dataset.sort_by_similarity(dataset.first().id, k=5)

view = dataset.sort_by_similarity("kites high in the sky", k=5)
```

### DIFF
--- a/fiftyone/brain/internal/core/mongodb.py
+++ b/fiftyone/brain/internal/core/mongodb.py
@@ -106,10 +106,10 @@ class MongoDBSimilarity(Similarity):
         # eg Atlas clusters generally have hostnames which end in "mongodb.net"
         # https://stackoverflow.com/q/73180110
         #
-        fou.ensure_package("pymongo>=4.5,<4.9")
+        fou.ensure_package("pymongo>=4.7,<4.9")
 
     def ensure_usage_requirements(self):
-        fou.ensure_package("pymongo>=4.5,<4.9")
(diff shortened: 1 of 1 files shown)


allenleetc (approved):
Code LGTM

Tested with grouped image dataset. Search is fixed when searching the original grouped dataset, both in-App or via SDK. Searching a flattened view (eg generated via `dataset.select_group_slices()` ) fails with a different, probably unrelated, error.

PR #219, opened 2024-12-17
Release v0.18.2

brimoor:


fiftyone/brain/internal/core/mongodb.py
```diff
@@ -106,10 +106,10 @@ def ensure_requirements(self):
         # eg Atlas clusters generally have hostnames which end in "mongodb.net"
         # https://stackoverflow.com/q/73180110
         #
-        fou.ensure_package("pymongo>=4.5,<4.9")
+        fou.ensure_package("pymongo>=4.7,<4.9")
```
findtopher, line 109: I don't think this will work, given that [fiftyone](https://github.com/voxel51/fiftyone/blob/release%2Fv1.2.0/setup.py#L58) now requires `"pymongo~=4.9.2"`

has this been tested with the current release branch?

brimoor: Tested with `pymongo==4.9.2` and it works. The `<4.9` was probably some kind of holdover from when this feature was initially launched (Atlas vector search was still in beta).

findtopher (approved):
🔩

PR #221, opened 2024-12-17
chore: Update svg to be fiftyone-brain

afoley587:
The current README shows the `fiftyone` version under `pypi package` (See below)

image

But, we should have it show the `fiftyone-brain` version, not the `fiftyone` version.

Screenshot 2024-12-17 at 11 36 22 AM


findtopher (approved):
slack.voxel51.com redirects to discord now... should we also update the Slack shield to be Discord?

PR #222, opened 2024-12-20
PGVector integration

xenophenes:
Hi!

These files still need to be polished up and the documentation + similarity test added. I'm opening this draft PR so existing work can be reviewed if desired and so the current progress is known. 

Additional changes will be pushed up to this PR ASAP.

The existing code will have a pre-commit hook performed on it before this PR is marked as ready to merge.

unnamed
unnamed1

### DIFF
--- a/fiftyone/brain/internal/core/pgvector.py
+++ b/fiftyone/brain/internal/core/pgvector.py
@@ -3,7 +3,6 @@ PGVector similarity backend.
 
 | Copyright 2017-2024, Voxel51, Inc.
 | `voxel51.com <https://voxel51.com/>`_
-|
 """
 import logging
 import numpy as np
@@ -33,6 +32,8 @@ class PgVectorSimilarityConfig(SimilarityConfig):
         ssl_key=None,
         ssl_root_cert=None,
         work_mem="64MB",  # Default work_mem value
+        index_name="embedding_hnsw_index",  # Default index name
+        embedding_column="embedding",  # Default embedding column name
         **kwargs,
     ):
         if metric not in _SUPPORTED_METRICS:
@@ -46,6 +47,8 @@ class PgVectorSimilarityConfig(SimilarityConfig):
         self.ssl_key = ssl_key
         self.ssl_root_cert = ssl_root_cert
         self.work_mem = work_mem
+        self.index_name = index_name  # User-specified index name
+        self.embedding_column = embedding_column  # User-specified embedding column name
 
     @property
     def method(self):
@@ -82,6 +85,7 @@ class PgVectorSimilarity(Similarity):
             samples, self.config, brain_key, backend=self
         )
 
+
 class PgVectorSimilarityIndex(SimilarityIndex):
     def __init__(self, samples, config, brain_key, backend=None):
         super().__init__(samples, config, brain_key, backend=backend)
@@ -112,7 +116,7 @@ class PgVectorSimilarityIndex(SimilarityIndex):
     def get_embeddings(self):
         try:
             logger.info("Fetching embeddings from the database...")
-            self._cur.execute("SELECT id, embedding FROM embeddings")
+            self._cur.execute(f"SELECT id, {self.config.embedding_column} FROM embeddings")
             return {row[0]: np.array(row[1]) for row in self._cur.fetchall()}
         except Exception as e:
             logger.error("Failed to fetch embeddings.")
@@ -132,12 +136,11 @@ class PgVectorSimilarityIndex(SimilarityIndex):
 
     def cleanup(self):
         try:
-            logger.info("Cleaning up: Deleting HNSW index and embeddings table...")
+            logger.info(f"Cleaning up: Deleting HNSW index '{self.config.index_name}' and embeddings table...")
 
             # Drop the HNSW index if it exists
-            index_name = "embedding_hnsw_index"
-            self._cur.execute(f"DROP INDEX IF EXISTS {index_name};")
-            logger.info(f"HNSW index '{index_name}' deleted successfully.")
+            self._cur.execute(f"DROP INDEX IF EXISTS {self.config.index_name};")
+            logger.info(f"HNSW index '{self.config.index_name}' deleted successfully.")
 
             # Optionally, drop the embeddings table
             self._cur.execute("DROP TABLE IF EXISTS embeddings;")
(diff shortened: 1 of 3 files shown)


fiftyone/brain/internal/core/pgvector.py
```diff
@@ -0,0 +1,142 @@
...
+                    embedding vector(%s)
+                )
+            ''', (len(embeddings[0]),))
+            
+            for embedding, id in zip(embeddings, sample_ids):
+                self._cur.execute(
+                    'INSERT INTO embeddings (id, embedding) VALUES (%s, %s::vector) ON CONFLICT (id) DO UPDATE SET embedding = EXCLUDED.embedding',
+                    (id, embedding.tolist())
+                )
+            self._conn.commit()
```
thesteve0, line 106: I am not sure if this is the right place but somewhere you need to create an HNSW index. I think the default values for index density are fine. You should probably ensure that working memory is temporarily set larger than the default. See here
https://github.com/thesteve0/arvix-query/blob/main/sql_commands.sql#L23

Cosine is also a fine distance metric to have as the default.

fiftyone/brain/internal/core/pgvector.py
```diff
@@ -0,0 +1,142 @@
...
+    SimilarityIndex,
+)
+import fiftyone.brain.internal.core.utils as fbu
+
+pgvector = fou.lazy_import("pgvector")
+logger = logging.getLogger(__name__)
+
+_SUPPORTED_METRICS = ("cosine", "dotproduct", "euclidean")
+
+class PgVectorSimilarityConfig(SimilarityConfig):
```
thesteve0, line 25: If you look at the elasticsearch integration they allow for defining the distance metric. We should do the same:
https://github.com/voxel51/fiftyone-brain/blob/e3300432b83adcb515a9179220c7f1f6979d1448/fiftyone/brain/internal/core/elasticsearch.py#L62

We will also need the SSL certs and other connection types that people use with PostgreSQL

fiftyone/brain/internal/core/pgvector.py
thesteve0: Looks like the elasticsearch has an explicit method for index creation
https://github.com/voxel51/fiftyone-brain/blob/e3300432b83adcb515a9179220c7f1f6979d1448/fiftyone/brain/internal/core/elasticsearch.py#L365

fiftyone/brain/internal/core/pgvector.py
thesteve0, line 1: Still need an implementation for 

get_embeddings - https://github.com/voxel51/fiftyone-brain/blob/e3300432b83adcb515a9179220c7f1f6979d1448/fiftyone/brain/internal/core/mongodb.py#L375

remove_from_index  - https://github.com/voxel51/fiftyone-brain/blob/e3300432b83adcb515a9179220c7f1f6979d1448/fiftyone/brain/internal/core/mongodb.py#L333

fiftyone/brain/internal/core/pgvector.py
illustris-admin: Have added a new method: 'create_hnsw_index' in 'pgvector.py' - Latest commit of 01/07/2025 includes this change

fiftyone/brain/internal/core/pgvector.py
illustris-admin: Have updated the method: '_initialize' in 'pgvector.py' to accomodate this functionality - Latest commit of 01/07/2025 includes this change

fiftyone/brain/internal/core/pgvector.py
illustris-admin: Have added the method: 'remove_from_index' in 'pgvector.py' to accomodate this functionality - Latest commit of 01/07/2025 includes this change

fiftyone/brain/internal/core/pgvector.py
```diff
@@ -0,0 +1,293 @@
...
+            raise ValueError(
+                f"Unsupported metric '{metric}'. Supported values are {_SUPPORTED_METRICS}"
+            )
+        super().__init__(**kwargs)
+        self._connection_string = connection_string
+        self.metric = metric
+        self.ssl_cert = ssl_cert
+        self.ssl_key = ssl_key
+        self.ssl_root_cert = ssl_root_cert
+        self.work_mem = work_mem
```
thesteve0, line 48: We are going to need to support the user specifying the index name since they may have already created the index outside of Fiftyone. A table could actually have more than 1 embedding column and therefore would require more than 1 index. They could also create 2 indices on the same embedding column (trying different HNSW parameters) so they need to specify which one to use here. 

We need to be able to use that pre-made index but also delete that index from a fiftyone call. Since we need to do that it would be good to allow the user to specify the name of the index they want created.

fiftyone/brain/internal/core/pgvector.py
```diff
@@ -0,0 +1,293 @@
...
+    def get_embeddings(self):
+        try:
+            logger.info("Fetching embeddings from the database...")
+            self._cur.execute("SELECT id, embedding FROM embeddings")
+            return {row[0]: np.array(row[1]) for row in self._cur.fetchall()}
+        except Exception as e:
+            logger.error("Failed to fetch embeddings.")
+            raise RuntimeError("Failed to fetch embeddings") from e
+
+    def remove_from_index(self, sample_ids):
```
thesteve0, line 121: See comment on init about indices

fiftyone/brain/internal/core/pgvector.py
```diff
@@ -0,0 +1,293 @@
...
+            for sample_id in sample_ids:
+                self._cur.execute("DELETE FROM embeddings WHERE id = %s", (sample_id,))
+            self._conn.commit()
+            logger.info("Samples removed successfully.")
+        except Exception as e:
+            logger.error("Error removing samples. Rolling back transaction.")
+            self._conn.rollback()
+            raise e
+
+    def cleanup(self):
```
thesteve0, line 133: see comment above in init about indices

fiftyone/brain/internal/core/pgvector.py
```diff
@@ -0,0 +1,293 @@
...
+            self._cur.execute("SELECT embedding FROM embeddings WHERE id = %s", (sample_id,))
+            result = self._cur.fetchone()
+            if result:
+                return np.array(result[0])
+            raise ValueError(f"Sample ID {sample_id} not found in the database.")
+        except Exception as e:
+            logger.error(f"Error fetching embedding for sample ID {sample_id}: {e}")
+            raise
+
+    def text_to_embedding(self, text):
```
thesteve0, line 249: I assume this is for multimodal embedding search. Is that correct?

fiftyone/brain/internal/core/pgvector.py
```diff
@@ -0,0 +1,293 @@
...
+                raise ValueError("Unsupported query type. Must be sample ID, text, or numeric vector.")
+
+            # Perform k-NN search
+            query_str = ",".join(map(str, query_embedding))
+            logger.info(f"Performing k-NN search with top-{k} neighbors...")
+
+            self._cur.execute(
+                f"""
+                SELECT id, embedding <-> ARRAY[{query_str}]::vector AS distance
+                FROM embeddings ORDER BY distance ASC LIMIT %s;
```
thesteve0, line 278: What happens here if they don't specify k? Perhaps we should check that they set it or set it to something like 20 by default

fiftyone/brain/internal/core/pgvector.py
illustris-admin: yes, that is correct

fiftyone/brain/internal/core/pgvector.py
illustris-admin: I can set it as a default, and allow for it to be passed as a parameter

fiftyone/brain/internal/core/pgvector.py
```diff
@@ -0,0 +1,304 @@
...
+            logger.info("Embeddings table deleted successfully.")
+
+            # Commit changes to the database
+            self._conn.commit()
+        except Exception as e:
+            logger.error("Error during cleanup operation. Rolling back transaction.")
+            self._conn.rollback()
+            raise e
+
+    def add_to_index(self, embeddings, sample_ids, m=16, ef_construction=64, **kwargs):
```
prernadh, line 156: Could you add in the following args to add_to_index- 

```
overwrite=True,
allow_existing=True,
warn_existing=False,
reload=True,
```

You can follow a pattern similar to the Qdrant integration [here](https://github.com/voxel51/fiftyone-brain/blob/8658fc8796172da95b9bec45a79b15b72cd6f201/fiftyone/brain/internal/core/qdrant.py#L302)

fiftyone/brain/internal/core/pgvector.py
```diff
@@ -0,0 +1,304 @@
...
+                self.config.connection_string, **ssl_options
+            )
+            self._cur = self._conn.cursor()
+        except Exception as e:
+            raise ValueError(
+                "Failed to connect to PostgreSQL database. "
+                "Check your connection string and ensure the database is running."
+            ) from e
+
+    def get_embeddings(self):
```
prernadh, line 116: This PR currently does not support embeddings for patches. Patches are image crops that are generated using bounding box labels. Could you please add this in as well? The qdrant integration can be a good template to follow for it

prernadh (changes requested):
Some comments to make this integration consistent with our other vector db integrations.

fiftyone/brain/internal/core/pgvector.py
```diff
@@ -0,0 +1,304 @@
...
+            logger.error(f"Error fetching embedding for sample ID {sample_id}: {e}")
+            raise
+
+    def text_to_embedding(self, text):
+        """Convert a text string to an embedding using the provided model."""
+        if not self.embedding_model:
+            raise ValueError("No embedding model provided for text-to-embedding conversion.")
+        return self.embedding_model.encode(text)
+
+    def _kneighbors(self, query, k=None, return_dists=False):
```
prernadh, line 261: Critical: You need to add in the `reverse` argument to _kneighbors. Otherwise `sort_by_similarity` fails

prernadh:
Sorting by similarity after creating the index is not working- Here is the code snippet I am using
```
import fiftyone as fo
import fiftyone.brain as fob
import fiftyone.zoo as foz

pgvector_index = fob.compute_similarity(
    dataset,
    model = model,
    brain_key="pgvector_index",
    backend="pgvector",
    connection_string="connection_string"
)

query_id = dataset.take(1).first().id

# Programmatically construct a view containing the 15 most similar images
view = dataset.sort_by_similarity(query_id, k=15, brain_key="pgvector_index")
```

Issue #227, opened 2025-01-10
use LanceDB with Object Storages


jgugglberger:
Hi,

I tried to build a similarity index with LanceDB as backend. At first a tried to set the URI parameter to a local path, and it worked fine. However, I want to use an object storage to save the index (azure object store in my case). According to LanceDB documentation [(here](https://lancedb.github.io/lancedb/guides/storage/#object-stores)) this should be possible.

I set the `uri` parameter in `compute_similarity()` to something like `az://lancedbstorage/tables`, but this ends up being converted to a local path (something like `/apath/az:/lancedbstorage/tables`).

I think the problem is because every URI gets normalized with `normalize_path`:
https://github.com/voxel51/fiftyone-brain/blob/b966ceec14899df74781cfa505b60597e546bec4/fiftyone/brain/internal/core/lancedb.py#L66

jgugglberger:
I was able to get around this error with following monkey patch:

```python
from fiftyone.brain.internal.core.lancedb import LanceDBSimilarityConfig

_SUPPORTED_METRICS = {
    "cosine": "cosine",
    "euclidean": "l2",
}
 
def patched_init(self, table_name=None, metric="cosine", uri="/tmp/lancedb", **kwargs):
    if metric not in _SUPPORTED_METRICS:
        raise ValueError(
            "Unsupported metric '%s'. Supported values are %s"
            % (metric, tuple(_SUPPORTED_METRICS.keys()))
        )
 
    super(LanceDBSimilarityConfig, self).__init__(**kwargs)
 
    self.table_name = table_name
    self.metric = metric
 
    if uri.startswith("az://"):
        self._uri = uri 
    else:
        self._uri = fos.normalize_path(uri)
 
LanceDBSimilarityConfig.__init__ = patched_init
```

brimoor:
Hi @jgugglberger 👋 

In order to support cloud paths, I think it's fine to just remove `normalize_path()` and expect users to provide `uri` that is already normalized.

Would you like to contribute a PR for the patch since you discovered this? 💪

jgugglberger:
Ok, I will make a PR

jgugglberger:
Opened [PR](https://github.com/voxel51/fiftyone-brain/pull/228). Will close the issue with this comment.

PR #228, opened 2025-01-10
remove normalize_path to support cloud object storage URIs

jgugglberger:
Follow up on [this issue](https://github.com/voxel51/fiftyone-brain/issues/227).


brimoor (approved):
LGTM, thanks! 🥇

PR #234, opened 2025-02-04
Postgres integration

prernadh:
Adding in PGVector integration

### DIFF
--- a/fiftyone/brain/internal/core/pgvector.py
+++ b/fiftyone/brain/internal/core/pgvector.py
@@ -1,0 +1,433 @@
+"""
+PGVector similarity backend.
+"""
+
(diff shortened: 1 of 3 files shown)


fiftyone/brain/internal/core/pgvector.py
```diff
@@ -0,0 +1,699 @@
...
+
+        if single_query:
+            ids = ids[0]
+            if return_dists:
+                dists = dists[0]
+
+        if close_conn:
+            self._cur.close()
+            self._conn.close()
+
```
manushreegangwar, line 637: You can use close_connections method

manushreegangwar (approved):

PR #236, opened 2025-02-21
Add `create_index` option to `compute_visualization()`

ritch:
## Change log

- Adds optional `create_index=True` and `points_field` arguments to `compute_visualization()` to allow users to request that the visualization points are stored directly on the samples of their dataset, which allows for optimized lasso queries in the App

## Image visualization with spatial index

```py
import numpy as np
import fiftyone as fo
import fiftyone.brain as fob

n = 50000

dataset = fo.Dataset()
dataset.add_samples([fo.Sample(filepath=f"image{i}.jpg") for i in range(n)])

points = np.random.randn(n, 2)

results = fob.compute_visualization(
    dataset,
    points=points,
    create_index=True,
    brain_key="points",
)

assert dataset.has_field("points")
assert "points" in dataset.list_indexes()
assert results.config.points_field == "points"
assert results.points.shape == (n, 2)

dataset.clear_cache()
results = dataset.load_brain_results("points")

assert results.config.points_field == "points"
assert results.points.shape == (n, 2)
```

## Patch visualization with spatial index

```py
import numpy as np

import fiftyone as fo
import fiftyone.zoo as foz
import fiftyone.brain as fob

n = dataset.count("ground_truth.detections")
points = np.random.randn(n, 2)

results = fob.compute_visualization(
    dataset,
    points=points,
    patches_field="ground_truth",
    create_index=True,
    brain_key="points",
)

assert dataset.has_field("ground_truth.detections.points")
# Unfortunately we can't make use of database indexes by default
assert "ground_truth.detections.points" not in dataset.list_indexes()
assert results.config.points_field == "points"
assert results.points.shape == (n, 2)

dataset.clear_cache()
results = dataset.load_brain_results("points")

assert results.config.points_field == "points"
assert results.points.shape == (n, 2)
```

Unfortunately this isn't supported... :(

```py
expr = {
    "points": {
        "$geoWithin": {
            "$polygon": [[-1, -1], [1, -1], [1, 1], [-1, 1]]
        }
    }
}
view = dataset.filter_labels("ground_truth", expr)
```

### DIFF
--- a/fiftyone/brain/visualization.py
+++ b/fiftyone/brain/visualization.py
@@ -450,7 +450,6 @@ class VisualizationResults(fob.BrainResults):
         If :meth:`use_view` has been called to restrict the index, this value
         may be larger than the current :meth:`index_size`.
         """
-        print(self.point_field)
         if self.point_field:
             return len(self._curr_view)
         return len(self.points)
--- a/fiftyone/brain/visualization.py
+++ b/fiftyone/brain/visualization.py
@@ -63,6 +63,7 @@ def compute_visualization(
     )
     points = None if point_field else points
     points_not_provided = points is None
+    point_field_index = f"{point_field}_index" if point_field else None
 
     if method == "manual" and points_not_provided:
         raise ValueError(
@@ -116,6 +117,7 @@ def compute_visualization(
         patches_field=patches_field,
         num_dims=num_dims,
         point_field=point_field,
+        point_field_index=point_field_index,
         **kwargs,
     )
 
@@ -703,6 +705,9 @@ class VisualizationConfig(fob.BrainMethodConfig):
         patches_field (None): the sample field defining the patches being
             analyzed, if any
         num_dims (2): the dimension of the visualization space
+        point_field (None): the name of a field in which to store the
+            visualization points, if desired
+        point_field_index (None): the name of the index field for the
     """
 
     def __init__(
@@ -713,6 +718,8 @@ class VisualizationConfig(fob.BrainMethodConfig):
         model_kwargs=None,
         patches_field=None,
         num_dims=2,
+        point_field=None,
+        point_field_index=None,
         **kwargs,
     ):
         if similarity_index is not None and not etau.is_str(similarity_index):
@@ -727,6 +734,8 @@ class VisualizationConfig(fob.BrainMethodConfig):
         self.model_kwargs = model_kwargs
         self.patches_field = patches_field
         self.num_dims = num_dims
+        self.point_field = point_field
+        self.point_field_index = point_field_index
         super().__init__(**kwargs)
 
     @property
@@ -747,6 +756,13 @@ class Visualization(fob.BrainMethod):
             fields.append(self.config.point_field)
 
         return fields
+
+    def cleanup(self, samples, key):
+        dataset = samples._dataset
+        point_field_index = self.config.point_field_index
+        has_index = point_field_index in dataset.list_indexes()
+        if has_index:
+            dataset.delete_index(self.config.point_field)
 
 
 class UMAPVisualizationConfig(VisualizationConfig):
--- a/fiftyone/brain/visualization.py
+++ b/fiftyone/brain/visualization.py
@@ -323,6 +323,8 @@ def _generate_spatial_index(
         points_field, fof.ListField, subfield=fof.FloatField
     )
 
+    points = points.astype(float)
+
     if create_index:
         min_val, max_val = points.min(), points.max()
         dataset.create_index([(points_field, "2d")], min=min_val, max=max_val)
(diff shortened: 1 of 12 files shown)


fiftyone/brain/visualization.py
```diff
@@ -266,15 +287,60 @@ def _get_dimension(points):
...
+    return None
+
+
+def _get_min_max(points):
+    return np.min(points, axis=0), np.max(points, axis=0)
+
+
+def _define_point_field(samples, point_field, min_val, max_val):
+    dataset = samples._dataset
+    dataset.add_sample_field(
```
kaixi-wang, line 314: I don't think we should be adding and creating an index on the sample collection in a private function in fiftyone-brain... It should start/be defined in fiftyone

jacobsela: I feel like this reinforces the fact that we need some notion of hidden fields.

kaixi-wang: yeah we do kinda in fiftyone... but it doesn't make sense to add hidden fields to samples outside of fiftyone.

ritch: I'm just following an existing pattern. See [here](https://github.com/voxel51/fiftyone-brain/blob/98573d28f355efbc3f0693af8c7a73f1ffcc8f3f/fiftyone/brain/internal/core/uniqueness.py#L147) and [here](https://github.com/voxel51/fiftyone-brain/blob/98573d28f355efbc3f0693af8c7a73f1ffcc8f3f/fiftyone/brain/internal/core/representativeness.py#L139)

ritch: The field should be hidden though - I'll add that before merging.

jacobsela: Why does it matter where it happens? We just need a way of storing data on an FO dataset without exposing it to the user. Shouldn't plugins/brain/whatever else be able to store data for future use?

kaixi-wang: no imo it completely matters when you start thinking out of the oss context and how this is maintained

kaixi-wang: also mongo has a limit to the number of indexes on a collection (64). The current advise to make things faster is create indexes, so people likely already have a fair number of indexes on label fields, combined indexes, etc. If run the visualization 30 times and it creates 30 indexes...

ritch: The API will have to handle adhoc schema updates. This isn't a new paradigm. I believe you are making a comment outside the scope of this PR. If there are patterns that are not being followed OTOH I'll gladly fix those.

fiftyone/brain/visualization.py
```diff
@@ -266,15 +287,60 @@ def _get_dimension(points):
...
+    )
+
+
+def _populate_point_field(samples, point_field, points, progress=True):
+    samples_and_points = zip(samples, points)
+    iterator = (
+        fou.ProgressBar(samples_and_points) if progress else samples_and_points
+    )
+    for sample, point in iterator:
+        sample.set_field(point_field, point.tolist())
```
ritch, line 328: Is there a more efficient way to do this?

fiftyone/brain/visualization.py
```diff
@@ -266,15 +289,68 @@ def _get_dimension(points):
...
+
+def _get_min_max(points):
+    flat_points = points.flatten()
+    min, max = float(np.min(flat_points)), float(np.max(flat_points))
+    return min, max
+
+
+def _define_point_field(samples, point_field, min_val, max_val):
+    dataset = samples._dataset
+    dataset.add_sample_field(
```
ritch, line 318: This needs to be a read only field and somehow hidden!

fiftyone/brain/visualization.py
ritch: `set_values()` perhaps

tests/intensive/test_visualization.py
```diff
@@ -277,6 +278,66 @@ def test_similarity_index():
...
+
+    # cleanup
+    dataset.delete_brain_run(brain_key)
+
+    assert not dataset.has_sample_field(point_field)
+    indexes = dataset.list_indexes()
+    assert index_name not in indexes
+
+
+def test_points_field_patches():
```
ritch, line 314: @brimoor this should cover the intended behavior, but this test is currently failing.

fiftyone/brain/visualization.py
```diff
@@ -565,6 +642,45 @@ def visualize(
             **kwargs,
         )
 
+    def index_points(self, points_field=None, create_index=True):
```
ritch, line 645: This should have a `progress` param similar to other methods that can take a while.

brimoor: Added ✅

ritch:
@brimoor is there any blockers for merging this?

PR #240, opened 2025-03-06
Adding limit to query

prernadh:
Currently Redis backend only returns 10 results - from redis vector search [documentation ](https://redis.io/docs/latest/develop/interact/search-and-query/advanced-concepts/vectors/#search-with-vectors)

> When performing a KNN vector search, you specify <top_k> nearest neighbors. However, the default Redis query LIMIT parameter (used for pagination) is 10. In order to get <top_k> returned results, you must also specify LIMIT 0 <top_k> in your search command


brimoor (approved):
Ah good catch! 🙇

PR #243, opened 2025-03-17
workflow updates

findtopher:
builds are failing, let's try some updates

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -13,7 +13,7 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     steps:
       - name: Clone fiftyone-brain
         uses: actions/checkout@v1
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -120,16 +120,15 @@ jobs:
   publish:
     needs: [build]
     if: startsWith(github.ref, 'refs/tags/v')
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     steps:
```
afoley587, line 124: We should possibly add a `setup python` reference here:

```yaml
      - name: Set up Python 3.11
        uses: actions/setup-python@v2
        with:
          python-version: '3.11'
```

afoley587 (changes requested):

.github/workflows/build.yml
findtopher: OK, if the 3.9 that comes on the image isn't good enough....

afoley587 (approved):

PR #244, opened 2025-03-17
chore: Publish should happen after test

afoley587:
# Rationale

The publishing of the package should happen after successful tests.

## Changes

Adds a `needs: test` entry for `publish`.


findtopher:
Brian and Ben have the history on this, so I'd like their input before we change it

benjaminpkane (approved):
Historically tests were unstable. LGTM!

PR #263, opened 2025-10-22
Add support for "prompt" similarity queries with Model instances in the current session

brimoor:
## Release notes

- similarity indexes now support prompt queries (only) in the current Python session when they are created via the `compute_similarity(..., model=<Model instance>)` syntax

## Tested by

```py
import fiftyone as fo
import fiftyone.zoo as foz
import fiftyone.brain as fob

index = fob.compute_similarity(dataset, model=model, brain_key="img_sim")
# WARNING: this index will not support prompt queries in the App or in future Python sessions. You can support this by providing the string name of a zoo model rather than a Model instance to compute_similarity(model=).

assert model.can_embed_prompts == True
assert index.supports_prompts == True
assert index.config.supports_prompts == None  # not stored

# NEW BEHAVIOR: this works (does not on `develop`)
_ = index.sort_by_similarity("kites high in the sky", k=5)
_ = dataset.sort_by_similarity("kites high in the sky", brain_key="img_sim", k=5)

# simulate starting a new session
dataset.clear_cache()

index = dataset.load_brain_results("img_sim")

assert index.supports_prompts == False  # now False
assert index.config.supports_prompts == None  # still None

# these no longer work
try:
    _ = index.sort_by_similarity("kites high in the sky", k=5)
except Exception as e:
    print(e)

# no longer works
try:
    _ = dataset.sort_by_similarity("kites high in the sky", brain_key="img_sim", k=5)
except Exception as e:
    print(e)
```

### DIFF
--- a/fiftyone/brain/similarity.py
+++ b/fiftyone/brain/similarity.py
@@ -97,13 +97,21 @@ def compute_similarity(
     if etau.is_str(model):
         _model_kwargs = model_kwargs or {}
         _model = foz.load_zoo_model(model, **_model_kwargs)
-        try:
-            supports_prompts = _model.can_embed_prompts
-        except:
-            supports_prompts = None
     else:
         _model = model
-        supports_prompts = None
+
+    try:
+        supports_prompts = _model.can_embed_prompts
+    except:
+        supports_prompts = False
+
+    if brain_key is not None and supports_prompts and not etau.is_str(model):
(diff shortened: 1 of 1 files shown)


manushreegangwar (approved):
LGTM! :rocket: 

Tested the behavior with single and multiple prompts.

PR #264, opened 2025-10-29
Make compute_visualization with TSNE compatible with scikit-learn versions 1.7.0+

sherpan:
TL;DR This PR renames 'n_iter' to 'max_iter'

when running  `fob.compute_visualization(ds, method='tsne', brain_key='tsne_test_171') `on scikit-learn version  1.7.1. The following error occurs 

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
    fob.compute_visualization(ds, method='tsne', brain_key='tsne_test_sdk_171')
    ~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/siddharthmehta/sid_foe/fot/lib/python3.13/site-packages/fiftyone/brain/__init__.py", line 540, in compute_visualization
    return fbv.compute_visualization(
           ~~~~~~~~~~~~~~~~~~~~~~~~~^
        samples,
        ^^^^^^^^
    ...<17 lines>...
        **kwargs,
        ^^^^^^^^^
    )
    ^
  File "/Users/siddharthmehta/sid_foe/fot/lib/python3.13/site-packages/fiftyone/brain/visualization.py", line 150, in compute_visualization
    points = brain_method.fit(embeddings)
  File "/Users/siddharthmehta/sid_foe/fot/lib/python3.13/site-packages/fiftyone/brain/visualization.py", line 1042, in fit
    _tsne = skm.TSNE(
        n_components=self.config.num_dims,
    ...<6 lines>...
        verbose=verbose,
    )
TypeError: TSNE.__init__() got an unexpected keyword argument 'n_iter'
```
when using an older version of scikit-learn (1.6.1), it works but we get this warning 

`/Users/siddharthmehta/sid_foe/fot/lib/python3.13/site-packages/sklearn/manifold/_t_sne.py:1164: FutureWarning: 'n_iter' was renamed to 'max_iter' in version 1.5 and will be removed in 1.7`

### DIFF
--- a/fiftyone/brain/visualization.py
+++ b/fiftyone/brain/visualization.py
@@ -8,7 +8,6 @@ Visualization interface.
 from copy import deepcopy
 import inspect
 import logging
-import functools
 from tabnanny import verbose
 from packaging import version
 
@@ -1044,18 +1043,13 @@ class TSNEVisualization(Visualization):
         verbose = 2 if self.config.verbose else 0
 
         sklearn_version = version.parse(sklearn.__version__)
-
         iter_param = (
             "max_iter"
             if sklearn_version >= version.parse("1.5.0")
             else "n_iter"
         )
 
-        tsne_partial = functools.partial(
-            skm.TSNE, **{iter_param: self.config.max_iters}
-        )
-
-        _tsne = tsne_partial(
+        _tsne = skm.TSNE(
             n_components=self.config.num_dims,
             perplexity=self.config.perplexity,
             learning_rate=self.config.learning_rate,
@@ -1063,6 +1057,7 @@ class TSNEVisualization(Visualization):
             init="pca",
             random_state=self.config.seed,
             verbose=verbose,
+            **{iter_param: self.config.max_iters}
         )
         return _tsne.fit_transform(embeddings)
 
--- a/fiftyone/brain/visualization.py
+++ b/fiftyone/brain/visualization.py
@@ -1045,7 +1045,7 @@ class TSNEVisualization(Visualization):
             learning_rate=self.config.learning_rate,
             metric=self.config.metric,
             init="pca",  # "random" or "pca"
-            n_iter=self.config.max_iters,
+            max_iter=self.config.max_iters,
             random_state=self.config.seed,
             verbose=verbose,
         )
--- a/fiftyone/brain/visualization.py
+++ b/fiftyone/brain/visualization.py
@@ -8,8 +8,12 @@ Visualization interface.
 from copy import deepcopy
 import inspect
 import logging
+import functools
+from tabnanny import verbose
+from packaging import version
 
 import numpy as np
+import sklearn
 import sklearn.decomposition as skd
 import sklearn.manifold as skm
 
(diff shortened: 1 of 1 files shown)


brimoor (commented):
@sherpan does `max_iter` work for `scikit-learn<1.5`?

```
/Users/siddharthmehta/sid_foe/fot/lib/python3.13/site-packages/sklearn/manifold/_t_sne.py:1164: FutureWarning: 'n_iter' was renamed to 'max_iter' in version 1.5 and will be removed in 1.7
```

Asking because we don't currently require any minimum version of `scikit-learn` in the `fiftyone` requirements:

https://github.com/voxel51/fiftyone/blob/a9395cc47f26a940ceb86e53ac4c7555299c1f26/setup.py#L65

So it would be proper for us to do one of:
1. Support both `n_iter` and `max_iter` in `fiftyone-brain` so we can keep our unrestricted `scikit-learn` requirement
2. Update our requirement to `scikit-learn>=1.5` in the main `fiftyone` repo

Looks like `scikit-learn==1.5` was [released in May 2024](https://pypi.org/project/scikit-learn/1.5.0). Ideally we'd have support going back a bit further for a dependency like this, as having strict version requirements can cause issues for some users that want to use `fiftyone` in conjunction with other scientific computing dependencies.

swheaton:
One option to support both that is fancier than full duplication of function call - use inspect or the scikit module itself if it has it, to get version. use functools.partial to bind either n_iter or max_iter, then call that partial with the rest of the params.

sherpan:
@brimoor confirmed my PR breaks when scikit-learn version < 1.5.0 

so do we think we should do option #1 with @swheaton 's suggestion?

manushreegangwar:
> @brimoor confirmed my PR breaks when scikit-learn version < 1.5.0
> 
> so do we think we should do option #1 with @swheaton 's suggestion?

@sherpan I would suggest going with 1. You can use `partial` as suggested or write a wrapper around `skm.TSNE` which checks the version and replaces `n_iter` with `max_iter` in kwargs for version >= 1.5

fiftyone/brain/visualization.py
```diff
@@ -1039,13 +1043,24 @@ def fit(self, embeddings):
...
+            "max_iter"
+            if sklearn_version >= version.parse("1.5.0")
+            else "n_iter"
+        )
+
+        tsne_partial = functools.partial(
+            skm.TSNE, **{iter_param: self.config.max_iters}
+        )
+
+        _tsne = tsne_partial(
```
brimoor: `partial()` is fancy but you can just do this, right?

```py
_tsne = skm.TSNE(..., **{iter_param: self.config.max_iters})
```

fiftyone/brain/visualization.py
```diff
@@ -8,8 +8,11 @@
 from copy import deepcopy
 import inspect
 import logging
+from tabnanny import verbose
```
brimoor: `tabnanny` appears to be for internal dev; remove it?

brimoor (commented):
I defer to @voxel51/ml-dev as code owners for final review + approval

manushreegangwar (approved):
LGTM :rocket: 

Tested tsne viz on version `1.7.2` on `cifar-10` dataset

PR #268, opened 2025-11-06
docs(RELEASING): AS-1101 document release process

afoley587:
The other day, we did a fiftyone-brain release. It makes sense to write down the steps that we took so other, newer, engineers can also understand the process. 

This PR adds a new `RELEASING.md`. It made sense to me to put them in a separate file because `CONTRIBUTING.md` is used by (potentially) external collaborators and the release process doesn't apply to them.


brimoor (approved):
LGTM

kevin-dimichel (approved):
Thank you Alex!

PR #270, opened 2025-11-24
Redaction method for sensitive media in datasets

neerajaabhyankar:
# Rationale

This feature gives users the ability to mask or blur sensitive information from their datasets, thereby enabling them to share their data wider.
Users can specify the detection field and label classes of their choice, as well as the redaction region (bounding_box/segementation_mask) and method.

_This method assumes that the label field & classes to be redacted already exist. For functionality that includes model application, we may maintain a plugin in FO Labs._

## Changes

Added a new brain method `create_redaction`, that operates on a collection of samples and accepts various arguments relating to the redaction method.

## Testing

- `test_interface.py::test_redaction` for checking a basic operation
- `test_redaction.py` for checking edge cases

## Notes

- If the provided `label_field` is not present in the sample collection, raises an error
- If the provided `label_classes` are not present in the sample/collection, nothing is redacted in the new media

## Related

Earlier version at https://github.com/neerajaabhyankar/redaction-plugin.
This will still be maintained, since it includes a single-operation `apply_model` + `create_redaction` capability. (ToDo: update the plugin to use this brain method.)

Presently, there does not exist a very high-quality face detection model that is worth recommending (the closest is https://huggingface.co/EsraaFouad/detr_fine_tune_face_detection_final but it also detects animal faces). Application via huggingface is easy; open to opinions on having it in the plugin v/s the brain method v/s entirely up to the user to bring their own model for flexibility in what they want to redact.

## JIRA

https://voxel51.atlassian.net/browse/ML-279

### DIFF
--- a/fiftyone/brain/__init__.py
+++ b/fiftyone/brain/__init__.py
@@ -988,7 +988,7 @@ def create_redaction(
         label_field: the name of the label field to process. Can be of type
             :class:`fiftyone.core.labels.Detection` or
             :class:`fiftyone.core.labels.Detections`
-        label_classes: a list or a comma-separated string of the label classes to redact, containing sensitive data
+        label_classes: a list of the label classes to redact, containing sensitive data
         redaction_type: the area in which to perform the redaction. Can be one of the following:
             "bounding_box": apply to the bounding box of the label class
             "segmentation_mask": apply to the segmentation mask of the label class
@@ -996,10 +996,9 @@ def create_redaction(
             "mask": mask the sensitive data
             "gaussian_blur": blur the sensitive data using a Gaussian blur
             "stack_blur": blur the sensitive data using a moving stack of colors
-            "pixelate": pixelate the sensitive data region
         redaction_field: the name of the field to store the redaction in.
             If None: the redaction will be stored in a new field with the name
-            "redacted_{label_field}_{label_classes}_{redaction_type}_{redaction_method}"
+            "redacted_{label_field}_{redaction_type}_{redaction_method}"
         force_recreate: whether to force the redaction to be recreated even if it already exists
         create_as_new_sample: whether to create a new sample with the redaction.
             If False: the path to the redacted media will be added
--- a/fiftyone/brain/__init__.py
+++ b/fiftyone/brain/__init__.py
@@ -971,7 +971,6 @@ def create_redaction(
     redaction_type="bounding_box",
     redaction_method="gaussian_blur",
     redaction_field=None,
-    force_recreate=False,
     progress=None,
 ):
     """Creates a redacted media file for the specified label classes in the specified label field.
@@ -980,8 +979,7 @@ def create_redaction(
     Replaces the regions (based on the redaction_type = bounding_box/segmentation_mask)
     with a blurred/masked image based on the redaction_method.
 
-    The redacted media is added as a new media field specified by the redaction_field
-    and also added to the tags of the sample for easy filtering.
+    The redacted media is added as a new media field specified by the redaction_field.
 
     Args:
         samples: a :class:`fiftyone.core.collections.SampleCollection`
@@ -999,7 +997,6 @@ def create_redaction(
         redaction_field: the name of the field to store the redaction in.
             If None: the redaction will be stored in a new field with the name
             "redacted_{label_field}_{redaction_type}_{redaction_method}"
-        force_recreate: whether to force the redaction to be recreated even if it already exists
         progress: whether to render a progress bar (True/False), use the
             default value ``fiftyone.config.show_progress_bars`` (None), or a
             progress callback function to invoke instead
@@ -1016,6 +1013,5 @@ def create_redaction(
         redaction_type,
         redaction_method,
         redaction_field,
-        force_recreate,
         progress,
     )
--- a/fiftyone/brain/__init__.py
+++ b/fiftyone/brain/__init__.py
@@ -972,6 +972,8 @@ def create_redaction(
     redaction_method="gaussian_blur",
     redaction_field=None,
     remote_redacted_media_dir=None,
+    num_workers=None,
+    skip_failures=True,
     progress=None,
 ):
     """Creates a redacted media file for the specified label classes in the specified label field.
@@ -1001,6 +1003,12 @@ def create_redaction(
         remote_redacted_media_dir: the remote directory to store the redacted media in.
             Only available with FiftyOne Teams.
             If None: the redacted media will not be uploaded to the cloud.
+        num_workers (None): the number of worker processes to use when processing
+            samples. If None, uses the default from ``map_samples``. Only applicable
+            if ``map_samples`` supports multiprocessing.
+        skip_failures (True): whether to gracefully continue without raising an
+            error if redaction cannot be performed for a sample. Only applicable
+            if ``map_samples`` supports this parameter.
         progress: whether to render a progress bar (True/False), use the
             default value ``fiftyone.config.show_progress_bars`` (None), or a
             progress callback function to invoke instead
@@ -1018,5 +1026,7 @@ def create_redaction(
         redaction_method,
         redaction_field,
         remote_redacted_media_dir,
+        num_workers,
+        skip_failures,
         progress,
     )
--- a/fiftyone/brain/__init__.py
+++ b/fiftyone/brain/__init__.py
@@ -1000,13 +1000,6 @@ def create_redaction(
             If None: the redaction will be stored in a new field with the name
             "redacted_{label_field}_{redaction_type}_{redaction_method}"
         force_recreate: whether to force the redaction to be recreated even if it already exists
-        create_as_new_sample: whether to create a new sample with the redaction.
-            If False: the path to the redacted media will be added
-                to the redaction_field of the original sample.
-            If True: a new sample will be created with the redaction;
-                the redaction_field will point to its ID
-                and the new sample's original_sample_id will point to the original sample
-            Note: The redaction_field will be unique function of the label, method & type args
         progress: whether to render a progress bar (True/False), use the
             default value ``fiftyone.config.show_progress_bars`` (None), or a
             progress callback function to invoke instead
--- a/fiftyone/brain/__init__.py
+++ b/fiftyone/brain/__init__.py
@@ -970,6 +970,7 @@ def create_redaction(
     label_classes=None,
     redaction_type="bounding_box",
     redaction_method="gaussian_blur",
+    redaction_field=None,
     force_recreate=False,
     create_as_new_sample=False,
     num_workers=None,
@@ -991,7 +992,7 @@ def create_redaction(
             :class:`fiftyone.core.labels.Polylines`,
             :class:`fiftyone.core.labels.Keypoints`, or
             :class:`fiftyone.core.labels.TemporalDetections`
-        label_classes: the comma-separated string of the list of label classes to redact, containing sensitive data
+        label_classes: a list or a comma-separated string of the label classes to redact, containing sensitive data
         redaction_type: the area in which to perform the redaction. Can be one of the following:
             "bounding_box": apply to the bounding box of the label class
             "segmentation_mask": apply to the segmentation mask of the label class
@@ -1000,6 +1001,9 @@ def create_redaction(
             "gaussian_blur": blur the sensitive data using a Gaussian blur
             "stack_blur": blur the sensitive data using a moving stack of colors
             "pixelate": pixelate the sensitive data region
+        redaction_field: the name of the field to store the redaction in.
+            If None: the redaction will be stored in a new field with the name
+            "redacted_{label_field}_{label_classes}_{redaction_type}_{redaction_method}"
         force_recreate: whether to force the redaction to be recreated even if it already exists
         create_as_new_sample: whether to create a new sample with the redaction.
             If False: the path to the redacted media will be added
@@ -1021,6 +1025,7 @@ def create_redaction(
         label_classes,
         redaction_type,
         redaction_method,
+        redaction_field,
         force_recreate,
         create_as_new_sample,
         num_workers,
--- a/fiftyone/brain/__init__.py
+++ b/fiftyone/brain/__init__.py
@@ -972,7 +972,6 @@ def create_redaction(
     redaction_method="gaussian_blur",
     redaction_field=None,
     force_recreate=False,
-    num_workers=None,
     progress=None,
 ):
     """Creates a redacted media file for the specified label classes in the specified label field.
@@ -987,10 +986,8 @@ def create_redaction(
     Args:
         samples: a :class:`fiftyone.core.collections.SampleCollection`
         label_field: the name of the label field to process. Can be of type
-            :class:`fiftyone.core.labels.Detections`,
-            :class:`fiftyone.core.labels.Polylines`,
-            :class:`fiftyone.core.labels.Keypoints`, or
-            :class:`fiftyone.core.labels.TemporalDetections`
+            :class:`fiftyone.core.labels.Detection` or
+            :class:`fiftyone.core.labels.Detections`
         label_classes: a list or a comma-separated string of the label classes to redact, containing sensitive data
         redaction_type: the area in which to perform the redaction. Can be one of the following:
             "bounding_box": apply to the bounding box of the label class
@@ -1011,7 +1008,6 @@ def create_redaction(
                 the redaction_field will point to its ID
                 and the new sample's original_sample_id will point to the original sample
             Note: The redaction_field will be unique function of the label, method & type args
-        num_workers: the number of workers to use when performing the redaction
         progress: whether to render a progress bar (True/False), use the
             default value ``fiftyone.config.show_progress_bars`` (None), or a
             progress callback function to invoke instead
@@ -1029,6 +1025,5 @@ def create_redaction(
         redaction_method,
         redaction_field,
         force_recreate,
-        num_workers,
         progress,
     )
--- a/fiftyone/brain/__init__.py
+++ b/fiftyone/brain/__init__.py
@@ -962,3 +962,67 @@ def compute_leaky_splits(
         skip_failures=skip_failures,
         progress=progress,
     )
+
+
+def create_redaction(
+    samples,
+    label_field=None,
+    label_classes=None,
+    redaction_type="bounding_box",
+    redaction_method="gaussian_blur",
+    force_recreate=False,
+    create_as_new_sample=False,
+    num_workers=None,
+    progress=None,
+):
+    """Creates a redacted media file for the specified label classes in the specified label field.
+
+    TODO: Add detailed docstring describing what redaction does.
+
+    .. note::
+
+        Runs of this method can be referenced later via brain key
+        ``redaction_field``.
+
+    Args:
+        samples: a :class:`fiftyone.core.collections.SampleCollection`
+        label_field: the name of the label field to process. Can be of type
+            :class:`fiftyone.core.labels.Detections`,
+            :class:`fiftyone.core.labels.Polylines`,
+            :class:`fiftyone.core.labels.Keypoints`, or
+            :class:`fiftyone.core.labels.TemporalDetections`
+        label_classes: the comma-separated string of the list of label classes to redact, containing sensitive data
+        redaction_type: the area in which to perform the redaction. Can be one of the following:
+            "bounding_box": apply to the bounding box of the label class
+            "segmentation_mask": apply to the segmentation mask of the label class
+        redaction_method: the method to use to perform the redaction. Can be one of the following:
+            "mask": mask the sensitive data
+            "gaussian_blur": blur the sensitive data using a Gaussian blur
+            "stack_blur": blur the sensitive data using a moving stack of colors
+            "pixelate": pixelate the sensitive data region
+        force_recreate: whether to force the redaction to be recreated even if it already exists
+        create_as_new_sample: whether to create a new sample with the redaction.
+            If False: the path to the redacted media will be added
+                to the redaction_field of the original sample.
+            If True: a new sample will be created with the redaction;
+                the redaction_field will point to its ID
+                and the new sample's original_sample_id will point to the original sample
+            Note: The redaction_field will be unique function of the label, method & type args
+        num_workers: the number of workers to use when performing the redaction
+        progress: whether to render a progress bar (True/False), use the
+            default value ``fiftyone.config.show_progress_bars`` (None), or a
+            progress callback function to invoke instead
+    """
+    import fiftyone.brain.internal.core.redaction as fbr
+
+    return fbr.create_redaction(
+        samples,
+        label_field,
+        label_classes,
+        redaction_type,
+        redaction_method,
+        force_recreate,
+        create_as_new_sample,
+        num_workers,
+        progress,
+    )
--- a/fiftyone/brain/__init__.py
+++ b/fiftyone/brain/__init__.py
@@ -971,6 +971,7 @@ def create_redaction(
     redaction_type="bounding_box",
     redaction_method="gaussian_blur",
     redaction_field=None,
+    remote_redacted_media_dir=None,
     progress=None,
 ):
     """Creates a redacted media file for the specified label classes in the specified label field.
@@ -997,6 +998,9 @@ def create_redaction(
         redaction_field: the name of the field to store the redaction in.
             If None: the redaction will be stored in a new field with the name
             "redacted_{label_field}_{redaction_type}_{redaction_method}"
+        remote_redacted_media_dir: the remote directory to store the redacted media in.
+            Only available with FiftyOne Teams.
+            If None: the redacted media will not be uploaded to the cloud.
         progress: whether to render a progress bar (True/False), use the
             default value ``fiftyone.config.show_progress_bars`` (None), or a
             progress callback function to invoke instead
@@ -1013,5 +1017,6 @@ def create_redaction(
         redaction_type,
         redaction_method,
         redaction_field,
+        remote_redacted_media_dir,
         progress,
     )
--- a/fiftyone/brain/__init__.py
+++ b/fiftyone/brain/__init__.py
@@ -976,13 +976,13 @@ def create_redaction(
     progress=None,
 ):
     """Creates a redacted media file for the specified label classes in the specified label field.
+    Requires the label field to already be present in the samples.
 
-    TODO: Add detailed docstring describing what redaction does.
+    Replaces the regions (based on the redaction_type = bounding_box/segmentation_mask)
+    with a blurred/masked image based on the redaction_method.
 
-    .. note::
-
-        Runs of this method can be referenced later via brain key
-        ``redaction_field``.
+    The redacted media is added as a new media field by default,
+    with an option to add it as a new sample.
 
     Args:
         samples: a :class:`fiftyone.core.collections.SampleCollection`
--- a/fiftyone/brain/__init__.py
+++ b/fiftyone/brain/__init__.py
@@ -972,7 +972,6 @@ def create_redaction(
     redaction_method="gaussian_blur",
     redaction_field=None,
     force_recreate=False,
-    create_as_new_sample=False,
     num_workers=None,
     progress=None,
 ):
@@ -982,8 +981,8 @@ def create_redaction(
     Replaces the regions (based on the redaction_type = bounding_box/segmentation_mask)
     with a blurred/masked image based on the redaction_method.
 
-    The redacted media is added as a new media field by default,
-    with an option to add it as a new sample.
+    The redacted media is added as a new media field specified by the redaction_field
+    and also added to the tags of the sample for easy filtering.
 
     Args:
         samples: a :class:`fiftyone.core.collections.SampleCollection`
@@ -1016,6 +1015,9 @@ def create_redaction(
         progress: whether to render a progress bar (True/False), use the
             default value ``fiftyone.config.show_progress_bars`` (None), or a
             progress callback function to invoke instead
+    Returns:
+        a :class:`fiftyone.brain.redaction.RedactionResults`
+        This has a method generate_redacted_dataset() to generate a new dataset with the redaction.
     """
     import fiftyone.brain.internal.core.redaction as fbr
 
@@ -1027,7 +1029,6 @@ def create_redaction(
         redaction_method,
         redaction_field,
         force_recreate,
-        create_as_new_sample,
         num_workers,
         progress,
     )
--- a/fiftyone/brain/internal/core/redaction.py
+++ b/fiftyone/brain/internal/core/redaction.py
@@ -44,7 +44,6 @@ def create_redaction(
     redaction_method,
     redaction_field,
     force_recreate,
-    create_as_new_sample,
     num_workers,
     progress,
 ):
@@ -69,7 +68,6 @@ def create_redaction(
         redaction_method,
         redaction_field,
         force_recreate,
-        create_as_new_sample,
     )
     brain_key = config.redaction_field
     brain_method = config.build()
@@ -80,8 +78,6 @@ def create_redaction(
     select_fields = [
         label_field,
         brain_key,
-        "original_sample_id",
-        "redacted_sample_ids",
     ]
     select_fields = [
         ff for ff in select_fields if samples.get_field(ff) is not None
@@ -89,7 +85,13 @@ def create_redaction(
     view = samples.select_fields(select_fields)
 
     logger.info("Computing redaction...")
-    for sample in view.iter_samples(progress=progress):
+    redaction_view = view.filter_labels(
+        f"{config.label_field}.detections",
+        F("label").is_in(config.label_classes),
+        only_matches=False,
+    )
+
+    for sample in redaction_view.iter_samples(progress=progress):
         brain_method.process_sample(sample)
         sample.save()
 
@@ -118,7 +120,6 @@ class RedactionConfig(fob.BrainMethodConfig):
         redaction_method,
         redaction_field,
         force_recreate,
-        create_as_new_sample,
         **kwargs,
     ):
         super().__init__(**kwargs)
@@ -128,7 +129,6 @@ class RedactionConfig(fob.BrainMethodConfig):
         self.redaction_method = redaction_method
         self.redaction_field = redaction_field
         self.force_recreate = force_recreate
-        self.create_as_new_sample = create_as_new_sample
         self.make_redaction_field()
 
     def make_redaction_field(self):
@@ -172,7 +172,6 @@ class Redaction(fob.BrainMethod):
         self.redaction_type = None
         self.redaction_method = None
         self.force_recreate = None
-        self.create_as_new_sample = None
         self.redaction_field = None
         self.processing_frames = False
 
@@ -192,7 +191,6 @@ class Redaction(fob.BrainMethod):
         self.redaction_method = self.config.redaction_method
         self.redaction_field = self.config.redaction_field
         self.force_recreate = self.config.force_recreate
-        self.create_as_new_sample = self.config.create_as_new_sample
         self.processing_frames = samples._is_frame_field(
             self.config.label_field
         )
@@ -202,34 +200,6 @@ class Redaction(fob.BrainMethod):
             sample.filepath,
             rel_output_dir=self.redaction_field,
         )
-        # if self.create_as_new_sample:
-        #     if (
-        #         ("redacted_sample_ids" in sample)
-        #         and (sample["redacted_sample_ids"] is not None)
-        #         and (self.redaction_field in sample["redacted_sample_ids"])
-        #         and (
-        #             os.path.exists(
-        #                 sample._dataset[
-        #                     sample["redacted_sample_ids"][self.redaction_field]
-        #                 ]["filepath"]
-        #             )
-        #         )
-        #     ):
-        #         logger.debug(
-        #             f"Redaction already exists for sample {sample.id} --> {sample['redacted_sample_ids'][self.redaction_field]}"
-        #         )
-        #         if self.force_recreate:
-        #             sample._dataset.delete_sample(
-        #                 sample["redacted_sample_ids"][self.redaction_field]
-        #             )
-        #             sample["redacted_sample_ids"].pop(self.redaction_field)
-        #         else:
-        #             return
-        #     if self.redaction_field in sample.tags:
-        #         logger.debug(
-        #             f"This is a redaction of sample {sample['original_sample_id']}"
-        #         )
-        #         return
 
         if (
             (f"{self.redaction_field}_filepath" in sample)
@@ -255,27 +225,6 @@ class Redaction(fob.BrainMethod):
             )
         else:
             self.redact_file_at(redacted_media_path, sample[self.label_field])
-
-        # if self.create_as_new_sample:
-        #     redacted_sample = sample.copy()
-        #     redacted_sample[self.label_field] = self.filter_detections(
-        #         sample[self.label_field], keep_matching=False
-        #     )
-        #     redacted_sample["filepath"] = redacted_media_path
-        #     if self.redaction_field not in redacted_sample.tags:
-        #         redacted_sample.tags.append(self.redaction_field)
-        #     redacted_sample["original_sample_id"] = sample.id
-        #     logger.debug(
-        #         f"Adding redacted sample at filepath: {redacted_media_path}"
-        #     )
-        #     sample._dataset.add_sample(redacted_sample)
-        #     if ("redacted_sample_ids" not in sample) or (
-        #         sample["redacted_sample_ids"] is None
-        #     ):
-        #         sample["redacted_sample_ids"] = {}
-        #     sample["redacted_sample_ids"][
-        #         self.redaction_field
-        #     ] = redacted_sample.id
 
         # add the redacted filepath to the sample
         if self.redaction_field not in sample.tags:
@@ -314,10 +263,10 @@ class Redaction(fob.BrainMethod):
                 )
             for frame, detections_object in zip(vp, detections_object_list):
                 redacted_image = self._redact_entire_image(frame)
-                redacted_image = self._apply_redaction_to_image(
+                frame = self._apply_redaction_to_image(
                     frame, redacted_image, detections_object
                 )
-                vp.write(redacted_image)
+                vp.write(frame)
         shutil.move(temp_redacted_path, redacted_path)
 
     def _redact_entire_image(self, image):
@@ -342,7 +291,7 @@ class Redaction(fob.BrainMethod):
     def _apply_redaction_to_image(
         self, image, redacted_image, detections_object
     ):
-        for detection in self.filter_detections(detections_object).detections:
+        for detection in detections_object.detections:
             x1, y1, x2, y2 = get_corners_from_bbox(
                 detection.bounding_box, image.shape
             )
@@ -361,53 +310,17 @@ class Redaction(fob.BrainMethod):
                 )
         return image
 
-    def filter_detections(
-        self, detections_object: fol.Detections, keep_matching: bool = True
-    ) -> fol.Detections:
-        """
-        Filters the detections object to keep only the detections that match the redaction labels.
-        Args:
-            detections_object: fol.Detections object
-            keep_matching: bool to keep matching detections (True) or remove matching detections (False)
-        Returns:
-            fo.Detections object
-        """
-        filtered_detections = fo.Detections()
-        if (not hasattr(detections_object, "detections")) or (
-            detections_object.detections is None
-        ):
-            return filtered_detections
-        for detection in detections_object.detections:
-            if keep_matching and (
-                detection.label
-                in self.label_classes  # pylint: disable=unsupported-membership-test
-            ):
-                filtered_detections.detections.append(detection)
-            elif not keep_matching and (
-                detection.label
-                not in self.label_classes  # pylint: disable=unsupported-membership-test
-            ):
-                filtered_detections.detections.append(detection)
-        return filtered_detections
-
     def get_fields(self, samples, brain_key):
         fields = [
             self.label_field,
             f"{brain_key}_filepath",
-            # "original_sample_id",
-            # "redacted_sample_ids",
         ]
-
-        # if samples._is_frame_field(self.label_field):
-        #     fields.append(samples._FRAMES_PREFIX + f"{brain_key}_filepath")
-
         return fields
 
     def cleanup(self, samples, brain_key):
         label_field = self.label_field
 
-        # samples._dataset.delete_sample_fields(brain_key, error_level=1)
-        samples._dataset.delete_sample_fields(
+        samples._dataset.delete_sample_field(
             f"{brain_key}_filepath", error_level=1
         )
 
@@ -454,25 +367,26 @@ class RedactionResults(fob.BrainResults):
 
         select_fields = self.backend.get_fields(self.samples, self.key)
         redacted_view = self.samples.select_fields(select_fields)
+        # exclude detetions of type label_field from the redacted dataset
+        redacted_view = redacted_view.filter_labels(
+            f"{self.backend.label_field}.detections",
+            ~F("label").is_in(self.backend.label_classes),
+            only_matches=False,
+        )
 
         for redacted_sample in redacted_view.iter_samples():
             redacted_sample["filepath"] = redacted_sample[
                 f"{self.key}_filepath"
             ]
-            redacted_sample[
-                self.backend.label_field
-            ] = self.backend.filter_detections(
-                redacted_sample[self.backend.label_field], keep_matching=False
-            )
-
             redacted_sample.tags.append(self.key)
             redacted_dataset.add_sample(redacted_sample)
 
-        redacted_dataset.delete_sample_fields(
+        # remove the redacted field from the redacted dataset
+        redacted_dataset.delete_sample_field(
             f"{self.key}_filepath", error_level=1
         )
-        logger.info(f"Redacted dataset {redacted_dataset.name} created")
 
+        logger.info(f"Redacted dataset {redacted_dataset.name} created")
         return redacted_dataset
 
 
--- a/fiftyone/brain/internal/core/redaction.py
+++ b/fiftyone/brain/internal/core/redaction.py
@@ -12,6 +12,8 @@ from typing import Tuple
 import numpy as np
 import cv2
 
+cv2.setNumThreads(1)
+
 from eta.core.video import VideoProcessor
 
 import fiftyone as fo
@@ -88,7 +90,9 @@ def create_redaction(
         only_matches=True,
     )
 
-    redaction_view.map_samples(brain_method.process_sample_fn, save=True)
+    _ = list(
+        redaction_view.map_samples(brain_method.process_sample_fn, save=True)
+    )
 
     logger.info(f"Adding redacted media field: {redacted_filepath_field}")
     if redacted_filepath_field not in samples.app_config.media_fields:
@@ -96,16 +100,20 @@ def create_redaction(
         samples.save()
 
     # upload the redacted media to the cloud
-    cloud_view = redaction_view.match(F("local_path").exists())
     remote_base_dirs = list(
-        set([os.path.dirname(f) for f in cloud_view.values("filepath")])
+        set(
+            [
+                os.path.dirname(f)
+                for f in redaction_view.values("filepath")
+                if not fos.is_local(f)
+            ]
+        )
     )
     for remote_base_dir in remote_base_dirs:
         fos.ensure_dir(remote_base_dir)
         fos.upload_media(
-            cloud_view.match(F("filepath").contains_str(remote_base_dir)),
-            remote_dir=config.redaction_field,
-            rel_dir=remote_base_dir,
+            redaction_view.match(F("filepath").contains_str(remote_base_dir)),
+            remote_dir=fos.join(remote_base_dir, config.redaction_field),
             media_field=redacted_filepath_field,
             update_filepaths=True,
             cache=False,
@@ -195,9 +203,15 @@ class Redaction(fob.BrainMethod):
         )
         fos.copy_file(local_media_path, redacted_media_path)
         if self.processing_frames:
-            self.redact_video_file_at(redacted_media_path, self.label_field)
+            self.redact_video_file_at(
+                redacted_media_path, sample[self.label_field]
+            )
         else:
-            self.redact_image_file_at(redacted_media_path, self.label_field)
+            self.redact_image_file_at(
+                redacted_media_path, sample[self.label_field]
+            )
+
+        sample[f"{self.redaction_field}_filepath"] = redacted_media_path
 
     def redact_image_file_at(self, redacted_path, detections_object):
         if not detections_object.detections:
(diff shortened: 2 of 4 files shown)


fiftyone/brain/internal/core/redaction.py
```diff
@@ -0,0 +1,492 @@
...
+
+
+logger = logging.getLogger(__name__)
+
+
+_ALLOWED_TYPES = (
+    fol.Detections,
+    fol.Polylines,
+    fol.Keypoints,
+    fol.TemporalDetections,
```
manushreegangwar: Remove since redaction doesn't support these labels. Also, add `fol.Detection`

fiftyone/brain/internal/core/redaction.py
```diff
@@ -0,0 +1,492 @@
...
+        "original_sample_id",
+        "redacted_sample_ids",
+    ]
+    select_fields = [
+        ff for ff in select_fields if samples.get_field(ff) is not None
+    ]
+    view = samples.select_fields(select_fields)
+
+    logger.info("Computing redaction...")
+    view._dataset.persistent = True
```
manushreegangwar: What's the reason for doing this?

fiftyone/brain/internal/core/redaction.py
```diff
@@ -0,0 +1,492 @@
...
+        else:
+            try:
+                self.label_classes = list(self.label_classes)
+            except:
+                raise ValueError(
+                    f"Invalid label classes type: {type(self.label_classes)}"
+                )
+        redaction_labels = "_".join(self.label_classes)
+        redaction_field = f"redacted_{self.label_field.replace('.', '_')}_{redaction_labels}_{self.redaction_type}_{self.redaction_method}"
+        self.redaction_field = redaction_field
```
manushreegangwar: I would suggest making the redaction field user configurable. You can see the config params using brain info.

fiftyone/brain/internal/core/redaction.py
```diff
@@ -0,0 +1,492 @@
...
+            elif self.redaction_type == "segmentation_mask" and mask is None:
+                logger.warning(f"No mask found for detection: {detection}")
+                continue
+            else:
+                raise ValueError(
+                    f"Unknown redaction type: {self.redaction_type}"
+                )
+        return image
+
+    def filter_detections(
```
manushreegangwar: See https://docs.voxel51.com/cheat_sheets/filtering_cheat_sheet.html#detections for filtering

fiftyone/brain/internal/core/redaction.py
```diff
@@ -0,0 +1,492 @@
...
+def _get_outpath(inpath, rel_output_dir):
+    dir_path = os.path.dirname(inpath)
+    # dir_path = fos.normalize_path(dir_path)
+    filename = os.path.basename(inpath)
+    dir_path = os.path.join(dir_path, rel_output_dir)
+    os.makedirs(dir_path, exist_ok=True)
+    return os.path.join(dir_path, filename)
+
+
+def read_video_frames(filepath):
```
manushreegangwar: See https://github.com/voxel51/eta/blob/develop/eta/core/video.py for video reader/writer

fiftyone/brain/__init__.py
```diff
@@ -962,3 +962,67 @@ def compute_leaky_splits(
...
+    with an option to add it as a new sample.
+
+    Args:
+        samples: a :class:`fiftyone.core.collections.SampleCollection`
+        label_field: the name of the label field to process. Can be of type
+            :class:`fiftyone.core.labels.Detections`,
+            :class:`fiftyone.core.labels.Polylines`,
+            :class:`fiftyone.core.labels.Keypoints`, or
+            :class:`fiftyone.core.labels.TemporalDetections`
+        label_classes: the comma-separated string of the list of label classes to redact, containing sensitive data
```
manushreegangwar: Why not support an iterable instead of comma-separated string?

fiftyone/brain/internal/core/redaction.py
```diff
@@ -0,0 +1,492 @@
...
+                for frame in sample.frames.values()
+            ]
+            self.redact_video_file_at(
+                redacted_media_path, detections_object_list
+            )
+        else:
+            self.redact_file_at(redacted_media_path, sample[self.label_field])
+
+        if self.create_as_new_sample:
+            redacted_sample = sample.copy()
```
manushreegangwar: This would copy all fields in `sample`. Redacted sample should only contain the required fields and filtered detections from `sample` https://docs.voxel51.com/user_guide/using_datasets.html?_gl=1*1vonm1t*_gcl_au*MTI3OTEyMjEwMy4xNzI2MTQ5NDk3#adding-samples-to-a-dataset

manushreegangwar (commented):
Just a couple of high-level comments and questions. Let's discuss https://docs.voxel51.com/brain.html#managing-brain-runs and options for tracking original vs redacted samples offline

fiftyone/brain/__init__.py
neerajaabhyankar: Good catch -- I forgot to update the docstring. Iterables are also supported.

fiftyone/brain/internal/core/redaction.py
neerajaabhyankar: > You can see the config params using brain info.

Oh right, makes sense then. I only did this for a 1:1 mapping between run and field.

fiftyone/brain/internal/core/redaction.py
neerajaabhyankar: Only so that the new fields/samples are saved (I'd done so in the plugin) .. although now that I think about it, this should be left to the user to do independently 👍

fiftyone/brain/internal/core/redaction.py
neerajaabhyankar: I've added support for `fol.TemporalDetections`.
Not sure whether `fol.Detection` and `fol.TemporalDetection` will work as it stands, but I'll test & edit this.

fiftyone/brain/internal/core/redaction.py
neerajaabhyankar: Replaced this with the `VideoProcessor`, although need to figure out the right `out_opts` for the ffmpeg reader, and how system-agnostic they are. The current one works, but changes the encoding and increases the file size.

fiftyone/brain/internal/core/redaction.py
```diff
@@ -0,0 +1,440 @@
...
+        self.label_type = samples._get_label_field_type(
+            self.config.label_field
+        )
+        self.label_classes = self.config.label_classes
+
+        self.redaction_type = self.config.redaction_type
+        self.redaction_method = self.config.redaction_method
+        self.redaction_field = self.config.redaction_field
+        self.force_recreate = self.config.force_recreate
+        self.processing_frames = samples._is_frame_field(
```
manushreegangwar:
```
self.label_field, self.processing_frames = samples._handle_frame_field(
            self.config.label_field
        )
```
Can you do this instead?

fiftyone/brain/internal/core/redaction.py
```diff
@@ -0,0 +1,440 @@
...
+        pass
+
+    def register_samples(self, samples):
+        self.label_field, _ = samples._handle_frame_field(
+            self.config.label_field
+        )
+        self.label_type = samples._get_label_field_type(
+            self.config.label_field
+        )
+        self.label_classes = self.config.label_classes
```
manushreegangwar: nit:`register_samples` in other brain methods is used for calling `samples.<method>`

manushreegangwar: Can you move the non-`samples.<method>` assignment elsewhere?

fiftyone/brain/internal/core/redaction.py
```diff
@@ -0,0 +1,440 @@
...
+    select_fields = [
+        ff for ff in select_fields if samples.get_field(ff) is not None
+    ]
+    view = samples.select_fields(select_fields)
+
+    logger.info("Computing redaction...")
+    redaction_view = view.filter_labels(
+        f"{config.label_field}.detections",
+        F("label").is_in(config.label_classes),
+        only_matches=False,
```
manushreegangwar: Why not process view with `only_matches = True`?

fiftyone/brain/internal/core/redaction.py
```diff
@@ -0,0 +1,440 @@
...
+    )
+
+    for sample in redaction_view.iter_samples(progress=progress):
+        brain_method.process_sample(sample)
+        sample.save()
+
+    logger.info(f"Adding redacted media field: {brain_key}_filepath")
+    if f"{brain_key}_filepath" not in view._dataset.app_config.media_fields:
+        view._dataset.app_config.media_fields.append(f"{brain_key}_filepath")
```
manushreegangwar: What not do this on `samples`? 

`samples.app_config.media_fields.append(...)`

fiftyone/brain/internal/core/redaction.py
```diff
@@ -0,0 +1,440 @@
...
+
+    view._dataset.save()
```
manushreegangwar: Is this needed? `sample.save()` should save the redacted filepath.

fiftyone/brain/internal/core/redaction.py
```diff
@@ -0,0 +1,440 @@
...
+    results = RedactionResults(samples, config, brain_key, brain_method)
+    brain_method.save_run_results(samples, brain_key, results)
+
+    logger.info("Redaction computation complete")
+    return results
+
+
+# @todo move to `fiftyone/brain/redaction.py`
+# Don't do this hastily; `get_brain_info()` on existing datasets has this
+# class's full path in it and may need migration
```
manushreegangwar: remove

fiftyone/brain/internal/core/redaction.py
```diff
@@ -0,0 +1,440 @@
...
+                raise ValueError(
+                    f"Invalid label classes type: {type(self.label_classes)}"
+                )
+        redaction_labels = "_".join(self.label_classes)
+        redaction_field_string = f"redacted_{self.label_field.replace('.', '_')}_{redaction_labels}_{self.redaction_type}_{self.redaction_method}"
+        self.redaction_field = (
+            redaction_field_string
+            if self.redaction_field is None
+            else self.redaction_field
+        )
```
manushreegangwar: This can result in a very long name when a large number of label classes are being redacted. I would suggest removing `{redaction_labels}`. Users can always get the redacted label names from `RedactionConfig`

fiftyone/brain/internal/core/redaction.py
```diff
@@ -0,0 +1,440 @@
...
+
+    @property
+    def type(self):
+        return "redaction"
+
+    @property
+    def method(self):
+        return self.redaction_field
```
manushreegangwar:
```suggestion
        return self.redaction_method
```

fiftyone/brain/internal/core/redaction.py
```diff
@@ -0,0 +1,440 @@
...
+                sample[f"{self.redaction_field}_filepath"] = None
+                sample.tags.remove(self.redaction_field)
+            else:
+                return
+
+        shutil.copy(sample.filepath, redacted_media_path)
+
+        if self.processing_frames:
+            detections_object_list = [
+                frame[self.label_field.replace("frames.", "")]
```
manushreegangwar: `samples._FRAME_PREFIX` need not be `frames.`

manushreegangwar: Have you tried `samples.values(self.label_field)`? It will probably be faster but you would have to modify this method to be `samples` level

fiftyone/brain/internal/core/redaction.py
```diff
@@ -0,0 +1,440 @@
...
+        if self.redaction_field not in sample.tags:
+            sample.tags.append(self.redaction_field)
+        sample[f"{self.redaction_field}_filepath"] = redacted_media_path
+
+        sample.save()
+        return
+
+    def redact_file_at(self, redacted_path, detections_object):
+        image = cv2.imread(redacted_path)
+        redacted_image = self._redact_entire_image(image)
```
manushreegangwar, line 236: Since your `redaction_view` contains all samples, you are computing redacted image even when it is not needed. Can you optimize this?

fiftyone/brain/internal/core/redaction.py
```diff
@@ -0,0 +1,440 @@
...
+                "-c:v",
+                "libopenh264",  # this increases the video quality
+            ],
+        ) as vp:
+            if vp.total_frame_count != len(detections_object_list):
+                raise ValueError(
+                    f"Number of video frames ({vp.total_frame_count}) does not match the number of detections ({len(detections_object_list)})"
+                )
+            for frame, detections_object in zip(vp, detections_object_list):
+                redacted_image = self._redact_entire_image(frame)
```
manushreegangwar: Similar optimization applies here

fiftyone/brain/internal/core/redaction.py
```diff
@@ -0,0 +1,440 @@
...
+            elif self.redaction_type == "segmentation_mask" and mask is None:
+                logger.warning(f"No mask found for detection: {detection}")
+                continue
+            else:
+                raise ValueError(
+                    f"Unknown redaction type: {self.redaction_type}"
+                )
+        return image
+
+    def get_fields(self, samples, brain_key):
```
manushreegangwar: `samples` is not used in the method

fiftyone/brain/internal/core/redaction.py
```diff
@@ -0,0 +1,440 @@
...
+        samples: the :class:`fiftyone.core.collections.SampleCollection` used
+        config: the :class:`RedactionConfig` used
+        brain_key: the brain key
+        backend (None): a :class:`Redaction` backend
+    """
+
+    def __init__(self, samples, config, brain_key, backend=None):
+        super().__init__(samples, config, brain_key, backend=backend)
+
+    def generate_redacted_dataset(self, name=None, overwrite=False):
```
manushreegangwar, line 364: 🎉

fiftyone/brain/internal/core/redaction.py
```diff
@@ -0,0 +1,440 @@
...
+            if name is not None
+            else f"{self.samples._dataset.name}_{self.samples.name}_redacted"
+        )
+        redacted_dataset = fo.Dataset(
+            name=redacted_dataset_name, overwrite=overwrite
+        )
+
+        select_fields = self.backend.get_fields(self.samples, self.key)
+        redacted_view = self.samples.select_fields(select_fields)
+        # exclude detetions of type label_field from the redacted dataset
```
manushreegangwar: nit: typo

fiftyone/brain/internal/core/redaction.py
```diff
@@ -0,0 +1,440 @@
...
+
+
+logger = logging.getLogger(__name__)
+
+
+_ALLOWED_TYPES = (
+    fol.Detection,
+    fol.Detections,
+    fol.TemporalDetection,
+    fol.TemporalDetections,
```
manushreegangwar: Does this work with temporal detections? How do you use temporal detections in videos?

fiftyone/brain/__init__.py
```diff
@@ -962,3 +962,73 @@ def compute_leaky_splits(
...
+            "redacted_{label_field}_{label_classes}_{redaction_type}_{redaction_method}"
+        force_recreate: whether to force the redaction to be recreated even if it already exists
+        create_as_new_sample: whether to create a new sample with the redaction.
+            If False: the path to the redacted media will be added
+                to the redaction_field of the original sample.
+            If True: a new sample will be created with the redaction;
+                the redaction_field will point to its ID
+                and the new sample's original_sample_id will point to the original sample
+            Note: The redaction_field will be unique function of the label, method & type args
+        num_workers: the number of workers to use when performing the redaction
```
manushreegangwar: not used

fiftyone/brain/internal/core/redaction.py
```diff
@@ -0,0 +1,440 @@
...
+        self.force_recreate = force_recreate
+        self.make_redaction_field()
+
+    def make_redaction_field(self):
+        if isinstance(self.label_classes, str):
+            self.label_classes = [
+                label_name.strip()
+                for label_name in self.label_classes.strip("[")
+                .strip("]")
+                .split(",")
```
manushreegangwar: I would suggest removing this since some models output classes with commas: `traffic signal, traffic light` can be a single class

fiftyone/brain/internal/core/redaction.py
```diff
@@ -0,0 +1,440 @@
...
+            x1, y1, x2, y2 = get_corners_from_bbox(
+                detection.bounding_box, image.shape
+            )
+            mask = detection.get_mask()
+            if self.redaction_type == "segmentation_mask" and mask is not None:
+                mask = fit_mask_to_bbox(mask, (y2 - y1, x2 - x1))
+                image[y1:y2, x1:x2][mask] = redacted_image[y1:y2, x1:x2][mask]
+            elif self.redaction_type == "bounding_box":
+                image[y1:y2, x1:x2] = redacted_image[y1:y2, x1:x2]
+            elif self.redaction_type == "segmentation_mask" and mask is None:
```
manushreegangwar: nit: can you combine this with the first if?

fiftyone/brain/__init__.py
```diff
@@ -962,3 +962,73 @@ def compute_leaky_splits(
...
+    The redacted media is added as a new media field specified by the redaction_field
+    and also added to the tags of the sample for easy filtering.
+
+    Args:
+        samples: a :class:`fiftyone.core.collections.SampleCollection`
+        label_field: the name of the label field to process. Can be of type
+            :class:`fiftyone.core.labels.Detections`,
+            :class:`fiftyone.core.labels.Polylines`,
+            :class:`fiftyone.core.labels.Keypoints`, or
+            :class:`fiftyone.core.labels.TemporalDetections`
```
manushreegangwar: remove

manushreegangwar (changes requested):
Thanks for addressing the previous comments and restructuring! 

I have some suggestions regarding optimization and fiftyone samples. Haven't tried running this yet. Will do that soon and provide more feedback.

fiftyone/brain/internal/core/redaction.py
neerajaabhyankar: It omits the samples that don't have these label classes.
We'd need to add them in the final result separately (since we'd like to keep the ones where we do not redact anything as well).

~The drawback with the current approach is that images are being copied even if no redaction is needed.
Open to changing this to point the `redaction_field` to the original `filepath` without copying in such cases.~
Update: optimizing this as you suggested

fiftyone/brain/internal/core/redaction.py
neerajaabhyankar: right, that has the same effect and is cleaner. replaced

fiftyone/brain/internal/core/redaction.py
neerajaabhyankar: replaced with `samples.save()`
(I guess I need to save the new media field addition? Can remove this if it isn't needed)

fiftyone/brain/internal/core/redaction.py
neerajaabhyankar: hmm, so we only accept a list of strings like `["traffic signal, traffic light", "car", "vehicle"]`?
I'll play around with the UI and see whether I can find a cleaner way out.

fiftyone/brain/internal/core/redaction.py
neerajaabhyankar: right, had initially thought this was a templated method

fiftyone/brain/internal/core/redaction.py
neerajaabhyankar: Sorry, realized this is not merely the Detections type for video :)

Removed (although, should I add support for this too?)

fiftyone/brain/internal/core/redaction.py
neerajaabhyankar: > samples._FRAME_PREFIX need not be frames

I see; the docs say this in a bunch of places e.g.
https://docs.voxel51.com/api/fiftyone.core.utils.html#fiftyone.core.utils.split_frame_fields

samples.values(self.label_field) looks great actually, I'll try to use this along with the other optimization refactor

fiftyone/brain/internal/core/redaction.py
neerajaabhyankar: made it so that we only accept a list as input
(found a way to do this through the UI if we want to)

fiftyone/brain/__init__.py
```diff
@@ -962,3 +962,68 @@ def compute_leaky_splits(
...
+            :class:`fiftyone.core.labels.Detections`
+        label_classes: a list of the label classes to redact, containing sensitive data
+        redaction_type: the area in which to perform the redaction. Can be one of the following:
+            "bounding_box": apply to the bounding box of the label class
+            "segmentation_mask": apply to the segmentation mask of the label class
+        redaction_method: the method to use to perform the redaction. Can be one of the following:
+            "mask": mask the sensitive data
+            "gaussian_blur": blur the sensitive data using a Gaussian blur
+            "stack_blur": blur the sensitive data using a moving stack of colors
+            "pixelate": pixelate the sensitive data region
```
manushreegangwar: not implemented

fiftyone/brain/internal/core/redaction.py
```diff
@@ -0,0 +1,448 @@
...
+            name
+            if name is not None
+            else f"{self.samples._dataset.name}_{self.samples.name}_redacted"
+        )
+        redacted_dataset = fo.Dataset(
+            name=redacted_dataset_name, overwrite=overwrite
+        )
+
+        select_fields = self.backend.get_fields(self.key)
+        redacted_view = self.samples.select_fields(select_fields)
```
manushreegangwar, line 380: Can you test this with a video dataset? This was throwing an error on `quickstart-video`

```
redacted_dataset = results.generate_redacted_dataset(
    name="vid_redacted_dataset", overwrite=True
)
```

fiftyone/brain/internal/core/redaction.py
```diff
@@ -0,0 +1,448 @@
...
+        only_matches=False,
+    )
+
+    if brain_method.processing_frames:
+        detections_object_list = samples.values(config.label_field)
+    else:
+        detections_object_list = [
+            sample[config.label_field]
+            for sample in redaction_view.iter_samples()
+        ]
```
manushreegangwar: Why not use `redaction_view.values(config.label_field)` here?

Since only `only_matches=True`, you will get an empty list of detections for samples with no detections.

fiftyone/brain/internal/core/redaction.py
```diff
@@ -0,0 +1,448 @@
...
+    x1 = int(bbox[0] * image_shape[1])
+    y1 = int(bbox[1] * image_shape[0])
+    x2 = int((bbox[0] + bbox[2]) * image_shape[1])
+    y2 = int((bbox[1] + bbox[3]) * image_shape[0])
+    return (x1, y1, x2, y2)
+
+
+def _get_outpath(inpath, rel_output_dir):
+    dir_path = os.path.dirname(inpath)
+    # dir_path = fos.normalize_path(dir_path)
```
manushreegangwar: remove

fiftyone/brain/internal/core/redaction.py
```diff
@@ -0,0 +1,448 @@
...
+        elif self.processing_frames and all(
+            [
+                len(frame_detections.detections) == 0
+                for frame_detections in sample_detections
+            ]
+        ):
+            redacted_media_path = sample.filepath
+            logger.debug(
+                f"No detections to be redacted for sample {sample.id}"
+            )
```
manushreegangwar: Instead of doing this to set `sample[f"{self.redaction_field}_filepath"] = redacted_media_path` later, why not set `sample[f"{self.redaction_field}_filepath"] = sample.filepath` for all samples and then only iterate over sample in `redaction_view` with `only_matches=True`?

fiftyone/brain/internal/core/redaction.py
manushreegangwar: This would be better: https://github.com/voxel51/fiftyone-brain/pull/270/changes#r2628186460

fiftyone/brain/internal/core/redaction.py
manushreegangwar: Can you add a check for `detections_object` to be a non-empty list inside this method instead? It is better to check here rather than rely on other methods to only call this method with non-empty `detections_object`

manushreegangwar (changes requested):
I tested this and have a couple of more comments.

fiftyone/brain/__init__.py
neerajaabhyankar: ahh thanks for the catch

fiftyone/brain/internal/core/redaction.py
neerajaabhyankar: Added an additional check here.

Although it's just a time-saving thing; on the fence about whether we need this.
Calling `redact_image_file_at(self, redacted_path, detections_object)` with an empty `detections_object` won't break -- it will create a redacted_image but not apply it anywhere / edit the original image.
Similarly for `redact_video_file_at` (although in this case making an exact copy of the image without redactions may also change it -- https://github.com/voxel51/fiftyone-brain/pull/270#discussion_r2582841942).
Keeping in the added check for now.

Still need to keep the one in the `process_sample` method since that dictates whether or not to create an object at `redacted_media_path` though.

fiftyone/brain/internal/core/redaction.py
neerajaabhyankar: hmm yeah, that sounds cleaner 🙂

fiftyone/brain/internal/core/redaction.py
neerajaabhyankar: didn't realize how handy the `.values` operator is 😄 thanks.

fiftyone/brain/internal/core/redaction.py
manushreegangwar: Since `redact_image_file_at` is a "public" method and users can call it without going through `process_sample`, it is best to add the check here. This is just a minor optimization and is indeed to save time -- only generate a full redacted image when needed :)

Re: _Still need to keep the one in the process_sample method since that dictates whether or not to create an object at redacted_media_path though._
You can keep both but if you choose to do [this](https://github.com/voxel51/fiftyone-brain/pull/270/changes#r2628186460) then I think you can get rid of the check in `process_sample`.

fiftyone/brain/internal/core/redaction.py
neerajaabhyankar: found the bug:
```
        self.label_field, self.processing_frames = samples._handle_frame_field(
            self.config.label_field
        )
```
was changing "frames.detections" to just "detections" --> changed it to `_, self.processing_frames = ...` instead.
`self.label_field` is now still "frames.detections".

fiftyone/brain/internal/core/redaction.py
neerajaabhyankar: yep 💯 realized that as I implemented it :) much cleaner with this

fiftyone/brain/internal/core/redaction.py
```diff
@@ -0,0 +1,430 @@
...
+    select_fields = [
+        ff for ff in select_fields if samples.get_field(ff) is not None
+    ]
+    view = samples.select_fields(select_fields)
+
+    logger.info("Computing redaction...")
+
+    for sample in view.iter_samples():
+        sample[f"{config.redaction_field}_filepath"] = sample.filepath
+        sample.save()
```
manushreegangwar: faster alternative:
```
view.set_values(f"{config.redaction_field}_filepath", view.values("filepath"))
```

fiftyone/brain/internal/core/redaction.py
```diff
@@ -0,0 +1,430 @@
...
+        # add the redacted filepath to the sample
+        if self.redaction_field not in sample.tags:
+            sample.tags.append(self.redaction_field)
+        sample[f"{self.redaction_field}_filepath"] = redacted_media_path
+
+        sample.save()
+        return
+
+    def redact_image_file_at(self, redacted_path, detections_object):
+        if len(detections_object.detections) == 0:
```
manushreegangwar: nit: `if not detections_object.detections` is better since:

detections_object.detections can be `None`. This can happen when you `apply_model` to a `view` and then call `dataset.values` on the whole dataset.

manushreegangwar: also applies to `redact_video_file_at`

fiftyone/brain/internal/core/redaction.py
manushreegangwar: Working as expected 🎉

fiftyone/brain/internal/core/redaction.py
```diff
@@ -0,0 +1,430 @@
...
+            ~F("label").is_in(self.backend.label_classes),
+            only_matches=False,
+        )
+
+        for redacted_sample in redacted_view.iter_samples():
+            redacted_sample["filepath"] = redacted_sample[
+                f"{self.key}_filepath"
+            ]
+            redacted_sample.tags.append(self.key)
+            redacted_dataset.add_sample(redacted_sample)
```
manushreegangwar: to avoid iteration over samples, you can use:

```
        if overwrite and redacted_dataset_name in fo.list_datasets():
            fo.delete_dataset(redacted_dataset_name)
        redacted_dataset = redacted_view.clone(name=redacted_dataset_name)
        redacted_dataset.set_values(
            "filepath", redacted_view.values(f"{self.key}_filepath")
        )
```

fiftyone/brain/__init__.py
```diff
@@ -962,3 +962,67 @@ def compute_leaky_splits(
...
+            If None: the redaction will be stored in a new field with the name
+            "redacted_{label_field}_{redaction_type}_{redaction_method}"
+        force_recreate: whether to force the redaction to be recreated even if it already exists
+        create_as_new_sample: whether to create a new sample with the redaction.
+            If False: the path to the redacted media will be added
+                to the redaction_field of the original sample.
+            If True: a new sample will be created with the redaction;
+                the redaction_field will point to its ID
+                and the new sample's original_sample_id will point to the original sample
+            Note: The redaction_field will be unique function of the label, method & type args
```
manushreegangwar: remove

tests/intensive/test_redaction.py
```diff
@@ -0,0 +1,162 @@
...
+        label_field="ground_truth",
+        label_classes=["person", "car"],
+        redaction_type="bounding_box",
+        redaction_method="stack_blur",
+        force_recreate=True,
+    )
+    new_modification_time = os.path.getmtime(
+        test_view.first()[expected_brain_key + "_filepath"]
+    )
+    assert new_modification_time > modification_time
```
manushreegangwar: this is failiing

fiftyone/brain/__init__.py
```diff
@@ -962,3 +962,67 @@ def compute_leaky_splits(
...
+            "bounding_box": apply to the bounding box of the label class
+            "segmentation_mask": apply to the segmentation mask of the label class
+        redaction_method: the method to use to perform the redaction. Can be one of the following:
+            "mask": mask the sensitive data
+            "gaussian_blur": blur the sensitive data using a Gaussian blur
+            "stack_blur": blur the sensitive data using a moving stack of colors
+        redaction_field: the name of the field to store the redaction in.
+            If None: the redaction will be stored in a new field with the name
+            "redacted_{label_field}_{redaction_type}_{redaction_method}"
+        force_recreate: whether to force the redaction to be recreated even if it already exists
```
manushreegangwar: you should revisit the `force_recreate` functionality since you are initializing the `redaction_field_filepath` to be the same as `filepath`

it is okay to discard this imo

manushreegangwar (changes requested):
Tested the new changes! Once you have addressed the final set of comments, it should be ready to merge 🛩️ 
 
Filtering for boxes overlapping redacted regions can be added as a follow-up PR once this is merged.

fiftyone/brain/internal/core/redaction.py
kaixi-wang: Why are you saving the same field value as another field? This should never be necessary... 
If you need to save a new filepath, just save the new filepath after you create it.

fiftyone/brain/internal/core/redaction.py
```diff
@@ -0,0 +1,430 @@
...
+
+    redaction_view = view.filter_labels(
+        f"{config.label_field}.detections",
+        F("label").is_in(config.label_classes),
+        only_matches=True,
+    )
+    detections_object_list = redaction_view.values(config.label_field)
+
+    for sample, sample_detections in zip(
+        redaction_view.iter_samples(progress=progress), detections_object_list
```
kaixi-wang: if you are using values (redaction_view.values(config.label_field)), then you should be using set_values()

alternatively, if using iter_samples (much slower), you shouldn't be using .values()

fiftyone/brain/internal/core/redaction.py
```diff
@@ -0,0 +1,430 @@
...
+    x2 = int((bbox[0] + bbox[2]) * image_shape[1])
+    y2 = int((bbox[1] + bbox[3]) * image_shape[0])
+    return (x1, y1, x2, y2)
+
+
+def _get_outpath(inpath, rel_output_dir):
+    dir_path = os.path.dirname(inpath)
+    filename = os.path.basename(inpath)
+    dir_path = os.path.join(dir_path, rel_output_dir)
+    os.makedirs(dir_path, exist_ok=True)
```
kaixi-wang: I don't think this will work since sample.filepath can be a cloud path in enterprise. You'll need to use fiftyone.core.storage methods to read and write properly

fiftyone/brain/__init__.py
```diff
@@ -962,3 +962,67 @@ def compute_leaky_splits(
...
+        create_as_new_sample: whether to create a new sample with the redaction.
```
kaixi-wang: Remove this since it isn't in the kwargs.

fiftyone/brain/internal/core/redaction.py
```diff
@@ -0,0 +1,430 @@
...
+    brain_method.ensure_requirements()
+    brain_method.register_run(samples, brain_key, cleanup=False)
+    brain_method.register_samples(samples)
+
+    select_fields = [
+        label_field,
+        brain_key,
+    ]
+    select_fields = [
+        ff for ff in select_fields if samples.get_field(ff) is not None
```
kaixi-wang: I'm confused what redaction_field is. Is it valid for it not to exist? Should you throw an error instead of just filtering it out?

fiftyone/brain/internal/core/redaction.py
```diff
@@ -0,0 +1,430 @@
...
+
+    config = RedactionConfig(
+        label_field,
+        label_classes,
+        redaction_type,
+        redaction_method,
+        redaction_field,
+        force_recreate,
+    )
+    brain_key = config.redaction_field
```
kaixi-wang, line 72: This is awkward as brain_key is normally not a field name as it needs to be unique so you wouldn't be able to run this method more than once (for example if you added samples later and wanted to redact parts in the new samples)

fiftyone/brain/internal/core/redaction.py
```diff
@@ -0,0 +1,430 @@
...
+            overwrite: whether to overwrite the redacted dataset if it already exists
+        Returns:
+            fo.Dataset
+        """
+        redacted_dataset_name = (
+            name
+            if name is not None
+            else f"{self.samples._dataset.name}_{self.samples.name}_redacted"
+        )
+        redacted_dataset = fo.Dataset(
```
kaixi-wang: We should not be creating new datasets like this in the results

fiftyone/brain/internal/core/redaction.py
```diff
@@ -0,0 +1,430 @@
...
+            and (os.path.exists(sample[f"{self.redaction_field}_filepath"]))
+        ):
+            logger.debug(f"Redaction already exists for sample {sample.id}")
+            if self.force_recreate:
+                sample[f"{self.redaction_field}_filepath"] = None
+                sample.tags.remove(self.redaction_field)
+            else:
+                return
+
+        shutil.copy(sample.filepath, redacted_media_path)
```
kaixi-wang: this won't work for enterprise

kaixi-wang (changes requested):

fiftyone/brain/internal/core/redaction.py
neerajaabhyankar: 1. thanks for bringing my attention here, this should be `f"{config.redaction_field}_filepath"` not config.redaction_path (a.k.a. `brain_key`)
2. `f"{config.redaction_field}_filepath"` is the target field to which we are saving the new redacted images.
3. If this exists, we may edit it; if it doesn't, we create it.
I added the field creation earlier, so that I can use `set_values`

fiftyone/brain/internal/core/redaction.py
neerajaabhyankar: @kaixi-wang we want to create redactions for all samples in the view.
However, if some of them don't have any sensitive detections, we skip iterating over them (and skip the creation of a new image too). We don't want the `f"{config.redaction_field}_filepath"` to be None for those, so setting the `filepath` as the default beforehand to make the code cleaner.

fiftyone/brain/internal/core/redaction.py
neerajaabhyankar: ahh, I was in fact wondering the same.
Is there a precedent in enterprise on creating new files? What's the canon way to do so -- `fos.copy_file(sample.filepath, redacted_media_path)`?

fiftyone/brain/internal/core/redaction.py
neerajaabhyankar: I see. Is there an example of exporting a new dataset that I can borrow conventions from?
Or should I get rid of this ability entirely?

For context: If one has data with sensitive content that they want to share separately, this exports it post redaction (also post getting rid of the sensitive detection.
I understand this can be done manually (basically replicating this code as needed), but lmk if you think a handy function is helpful.

fiftyone/brain/__init__.py
neerajaabhyankar: > it is okay to discard this imo

I agree -- this doesn't seem to be standard in other brain methods. Should also address https://github.com/voxel51/fiftyone-brain/pull/270/files#r2636592325.

If a redaction already exists, should I default to recreating it always? This will also get rid of the if-conditions checking for the same.

neerajaabhyankar: On a related note: is the appending `redaction_field` to `sample.tags` functionality still useful? I don't see it being standard practice either.

fiftyone/brain/internal/core/redaction.py
neerajaabhyankar: I see; I found this to be set to the respective method's field (which is passed as a kwarg) everywhere else.

```
fiftyone/brain/internal/core/mistakenness.py:    brain_key = mistakenness_field
fiftyone/brain/internal/core/representativeness.py:    brain_key = representativeness_field
fiftyone/brain/internal/core/hardness.py:    brain_key = hardness_field
fiftyone/brain/internal/core/uniqueness.py:    brain_key = uniqueness_field
```

Option 1: I allow the user to call this method with the same `redaction_field` and only redact samples for which a redaction doesn't already exist. I also get rid of the `force_recreate` functionality and default to create-redaction-only-if-it-doesn't-exist.
Option 2: I append a uuid to the redaction_field and set that as the brain key (although this isn't done elsewhere).

What do you suggest?

fiftyone/brain/internal/core/redaction.py
manushreegangwar: Some images don't have any redactions applied so they are the same as  the original. While I understand that `filepath` can be used for samples where `redaction_filepath` is set to None (if not initially cloning the values), what is the issue with cloning the values? Redacted images are saved to a new filepath.

fiftyone/brain/internal/core/redaction.py
neerajaabhyankar: Got it. I'll address this in a bit, once I figure out a couple other things (for now added a ToDo here).

Update: used `map_samples` instead; [this suggests](https://docs.voxel51.com/user_guide/using_datasets.html#map-operations) it is just as fast.

**Another Update: Couldn't get `map_samples` to work as expected!**
I'll give it another shot tonight/tomorrow. Added a ToDo for now.

fiftyone/brain/internal/core/redaction.py
neerajaabhyankar: Thanks, looking into this.
I see pathlib being used in [some places](https://github.com/voxel51/fiftyone-teams/blob/main/docs/scripts/generate_plugin_docs.py#L40), but I'm testing this and options using `fos` with a cloud path before committing a change.

tests/intensive/test_redaction.py
neerajaabhyankar: might get rid of this test once I remove the `force_recreate` (leaving it commented out temporarily)

neerajaabhyankar (commented):
Thanks for the comments Kacey and Manushree!
I'm not done addressing them all yet, will push a part 2 soon, after some of the open questions are resolved.

fiftyone/brain/__init__.py
manushreegangwar: Yes, default to recreating and you can remove the `tags` since it is no longer useful.

fiftyone/brain/internal/core/redaction.py
manushreegangwar: @neerajaabhyankar You should keep the existing pattern. 

You can run this method with the same `brain_key` multiple times as long as your `label_field` remains the same for each run. You can verfiy this -- for example, by running on two complementary views of the same dataset with the same label field.

How your run behaves when the key exists is controlled by: https://github.com/voxel51/fiftyone/blob/develop/fiftyone/core/runs.py#L377

fiftyone/brain/internal/core/redaction.py
```diff
@@ -0,0 +1,413 @@
...
+        fos.ensure_dir(remote_base_dir)
+        fos.upload_media(
+            cloud_view.match(F("filepath").contains_str(remote_base_dir)),
+            remote_dir=config.redaction_field,
+            rel_dir=remote_base_dir,
+            media_field=redacted_filepath_field,
+            update_filepaths=True,
+            cache=False,
+            overwrite=True,
+        )
```
neerajaabhyankar: (still to be tested)

fiftyone/brain/internal/core/redaction.py
neerajaabhyankar: I'm using the local paths to create the redactions (going off of https://docs.voxel51.com/enterprise/cloud_media.html#cloud-media-caching).

Then, I am uploading media (as per https://docs.voxel51.com/enterprise/cloud_media.html#fiftyone-core-storage) relative to the base dirs of existing samples.

Looks a bit hacky to me, open to suggestions!
~Trying to test with enterprise.~ Update: tested with enterprise ✅
Still, open to any suggestions here.

fiftyone/brain/internal/core/redaction.py
neerajaabhyankar: Resolved by `cv2.setNumThreads(1)`; runs quite fast now!

Do let me know if there's a more idomatic syntax than `_ = list(redaction_view.map_samples(...))`

tests/intensive/test_redaction.py
```diff
@@ -30,7 +31,11 @@ def test_create_redaction_fields():
     assert brain_key in dataset.list_brain_runs()
     redacted_image_path = test_view.first()[brain_key + "_filepath"]
     assert redacted_image_path is not None
-    assert os.path.exists(redacted_image_path)
+    assert fos.exists(redacted_image_path)
+    assert (
+        test_view.first()[brain_key + "_filepath"]
+        != test_view.first()["filepath"]
+    )  # only since the first sample has >0 detections of label classes
```
manushreegangwar: Is this test based on first sample having at least one of the `label_classes`?

This may break if quickstart zoo dataset is updated. You can instead find samples in `test_view` where one of the `label_classes` are present and then make this assertion on those samples.

fiftyone/brain/internal/core/redaction.py
```diff
@@ -0,0 +1,427 @@
...
+    for remote_base_dir in remote_base_dirs:
+        fos.ensure_dir(remote_base_dir)
+        fos.upload_media(
+            redaction_view.match(F("filepath").contains_str(remote_base_dir)),
+            remote_dir=fos.join(remote_base_dir, config.redaction_field),
+            media_field=redacted_filepath_field,
+            update_filepaths=True,
+            cache=False,
+            overwrite=True,
+        )
```
manushreegangwar: @kaixi-wang could you talke a look at this?

Does `map_samples` speed things up significantly, @neerajaabhyankar?

manushreegangwar (commented):
Have you tested the latest changes on both FOE and OSS, @neerajaabhyankar?

fiftyone/brain/internal/core/redaction.py
```diff
@@ -0,0 +1,427 @@
...
+            [
+                os.path.dirname(f)
+                for f in redaction_view.values("filepath")
+                if not fos.is_local(f)
+            ]
+        )
+    )
+    for remote_base_dir in remote_base_dirs:
+        fos.ensure_dir(remote_base_dir)
+        fos.upload_media(
```
kaixi-wang: Is there a precedent for this pattern? Otherwise, I feel like we should not be inferring directories and saving files to cloud storage without users explicitly providing a cloud path.

fiftyone/brain/internal/core/redaction.py
```diff
@@ -0,0 +1,427 @@
...
+    if redacted_filepath_field not in samples.app_config.media_fields:
+        samples.app_config.media_fields.append(redacted_filepath_field)
+        samples.save()
+
+    # upload the redacted media to the cloud
+    remote_base_dirs = list(
+        set(
+            [
+                os.path.dirname(f)
+                for f in redaction_view.values("filepath")
```
kaixi-wang: this does not scale. There are very few cases where you should ever have more than 1 values call for the same field in a single function.

fiftyone/brain/internal/core/redaction.py
```diff
@@ -0,0 +1,427 @@
...
+        )
+
+    results = RedactionResults(samples, config, brain_key, brain_method)
+    brain_method.save_run_results(samples, brain_key, results)
+
+    logger.info("Redaction computation complete")
+    return results
+
+
+class RedactionConfig(fob.BrainMethodConfig):
```
kaixi-wang, line 138: if you add the new media field dir as a required arg, you would avoid needing to check every filepath for is_local and you aren't potentially messing up users cloud buckets organization.

kaixi-wang (changes requested):

tests/intensive/test_redaction.py
neerajaabhyankar: Oh if it can change, then yes I'll robustify it.

Re. FOE, yes I tested it with a local setup + test script. Should I add that to the repo as well? I don't see an explicit Teams path in the CI tests.

fiftyone/brain/internal/core/redaction.py
neerajaabhyankar: It did speed up by more than 5x consistently :)

fiftyone/brain/internal/core/redaction.py
neerajaabhyankar: I don't see a precedent for uploading files at all in FOB.

I agree, asking for a storage dir makes sense.
(Although it's possible that the original data is scattered across multiple dirs, and the redacted one now lives in a single place, which seems less elegant/flexible) but if it makes things more predictable for the user, should be okay.

fiftyone/brain/internal/core/redaction.py
neerajaabhyankar: Replacing the target dir with a user-provided arg will remove the need for this.

fiftyone/brain/internal/core/redaction.py
neerajaabhyankar: ~As for whether the user-provided path is a valid one with write access, I'll add a check at the beginning before we do any redaction.~

Update: I think we should make it optional -- should be possible to run redaction without uploading to cloud as well.

neerajaabhyankar: Done; a couple comments:

- I see a [UniqueFilenameMaker](https://github.com/voxel51/fiftyone-teams/blob/develop/fiftyone/core/utils.py#L1946) which I'm guessing prevents two media with `dir1/img0.png` and `dir2/img0.png` from clashing. I haven't tested this, but can if necessary.
- Likewise, I found `fos.ensure_dir(remote_redacted_media_dir)` redundant when I tested this, but if it isn't, I can add it before the upload.
- `hasattr(fos, "upload_media")` seems like a hacky way of checking whether or not I'm in teams -- lmk if there's a better way.

neerajaabhyankar: @kaixi-wang does this look okay?

manushreegangwar (approved):
LGTM! :rocket:

fiftyone/brain/internal/core/redaction.py
```diff
@@ -0,0 +1,426 @@
...
+"""
+import os
+import logging
+import shutil
+from typing import Tuple
+
+import numpy as np
+import cv2
+
+cv2.setNumThreads(1)
```
kaixi-wang: wouldn't this affect other functions? Do you know if this is beneficial in all cases and platforms? I recall @PeterOttoMoeller had an issue with multiprocessing performance and setting the torch threads might have helped, but it was not clear why/if it would hurt performance for others

kaixi-wang:
@coderabbitai full review

fiftyone/brain/internal/core/redaction.py
```diff
@@ -0,0 +1,426 @@
...
+        return image
+
+    def get_fields(self, brain_key):
+        fields = [
+            self.label_field,
+            f"{brain_key}_filepath",
+        ]
+        return fields
+
+    def cleanup(self, samples, brain_key):
```
kaixi-wang, line 327: The cleanup here doesn't delete the media, just the fields?

fiftyone/brain/internal/core/redaction.py
```diff
@@ -0,0 +1,426 @@
...
+        samples: the :class:`fiftyone.core.collections.SampleCollection` used
+        config: the :class:`RedactionConfig` used
+        brain_key: the brain key
+        backend (None): a :class:`Redaction` backend
+    """
+
+    def __init__(self, samples, config, brain_key, backend=None):
+        super().__init__(samples, config, brain_key, backend=backend)
+
+    def generate_redacted_dataset(self, name=None, overwrite=False):
```
kaixi-wang, line 364: this generates a non persistent dataset that won't be visible in teams. Is that intended or should you add persistent as a kwarg?

fiftyone/brain/internal/core/redaction.py
```diff
@@ -0,0 +1,426 @@
...
+
+    view.set_values(redacted_filepath_field, view.values("filepath"))
+    redaction_view = view.filter_labels(
+        f"{config.label_field}.detections",
+        F("label").is_in(config.label_classes),
+        only_matches=True,
+    )
+
+    _ = list(
+        redaction_view.map_samples(brain_method.process_sample_fn, save=True)
```
kaixi-wang: maybe expose a num_workers kwarg to customize the number of processes?

fiftyone/brain/internal/core/redaction.py
```diff
@@ -0,0 +1,426 @@
...
+            logger.info(
+                f"Uploading redacted media to {remote_redacted_media_dir} ..."
+            )
+            fos.upload_media(
+                redaction_view,
+                remote_dir=remote_redacted_media_dir,
+                media_field=redacted_filepath_field,
+                update_filepaths=True,
+                cache=False,
+                overwrite=True,
```
kaixi-wang, line 124: should we also expose overwriting in the kwarg?

tests/intensive/test_redaction.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+def test_create_redaction_samples_video():
+    dataset = foz.load_zoo_dataset("quickstart-video").clone()
+    test_view = dataset.take(1, seed=42)
+
+    brain_key = "test_create_redaction_samples_video"
+    results = fob.create_redaction(
+        test_view,
+        label_field="frames.detections",
```
kaixi-wang: how is this test passing? I don't think frames.detections is a field on quickstart-video?

kaixi-wang (commented):
I just tried running this
```
import fiftyone as fo
import fiftyone.brain as fob
ds=fo.load_dataset('quickstart')
classes = ds.distinct("ground_truth.detections.label")[:5]
print('Redacting the following classes:', classes)
brain_key = "test_create_redaction_samples_video"
results = fob.create_redaction(
    ds,
    label_field="ground_truth",
    label_classes=classes,
    redaction_type="bounding_box",
    redaction_method="mask",
    redaction_field=brain_key,
)
redacted_dataset = results.generate_redacted_dataset(
    name="redacted-qs-gt", overwrite=True
)
redacted_dataset.persistent=True
print(redacted_dataset)
```

However, the redacted_dataset has the same filepaths as quickstart?

fiftyone/brain/internal/core/redaction.py
neerajaabhyankar: Can do that; although I want to make sure what happens if these paths exist in the cloud and overwrite=False; I assume
- an exception is raised
- all the previous steps (redaction + local files being saved) still go through
- now the onus is on the user to call `upload_media` again if they want to?

Seems a bit complicated, although if our users are used to dealing with things like these, sounds okay to me.

fiftyone/brain/internal/core/redaction.py
neerajaabhyankar: Yep, that's what it was doing, although I somehow never revisited that decision.

Makes sense to clean up the files too, adding it, thanks :)

fiftyone/brain/internal/core/redaction.py
neerajaabhyankar: hmm, I tested this with teams I do see the resultant dataset.
When this method returns a dataset, the user can choose to make it persistent if they want to, right?

Here's my testing script
[test_redaction_teams.py](https://github.com/user-attachments/files/24682386/test_redaction_teams.py)

tests/intensive/test_redaction.py
neerajaabhyankar: yeah it doesn't show up in the schema, but [this](https://docs.voxel51.com/user_guide/using_aggregations.html#aggregating-frame-labels) suggests that `frames.blah` is the right replacement for label_fields in video datasets.

Do you have a stack trace for `pytest tests/intensive/test_interface.py -s -k test_redaction_video`?

neerajaabhyankar: hmm, I think the test was passing vacuously (although I do remember "frames.detections" working before in this context!)
Fixed it by using the frame_label_field from
```
        self.frame_label_field, self.processing_frames = samples._handle_frame_field(
            self.label_field
        )
```

fiftyone/brain/internal/core/redaction.py
neerajaabhyankar: I moved this inside the `_redact_entire_image` function (which is where the cv2 processes were hanging).
Works now, afaik this would be safe, although do correct me if I'm wrong :)

PR #272, opened 2025-12-01
Ensure ids exist in lancedb index when querying

ehofesmann:
# Rationale

There exist a few bugs in the lancedb integration when attempting to perform similarity search queries if:

- You query on a view that includes samples not in the lancedb table (for example, when querying on a slice of a grouped dataset)
```python
import fiftyone.brain as fob
import fiftyone.zoo as foz
import fiftyone as fo
import numpy as np

dataset = foz.load_zoo_dataset("quickstart-groups")
embeddings = np.random.random((100, 512))
dataset[:100].set_values("embeddings", embeddings)
index = fob.compute_similarity(
    dataset.exists("embeddings"),
    embeddings="embeddings",
    brain_key="group_sim",
    backend="lancedb",
)

session = fo.launch_app(dataset)
# SelectGroupSlices for the `left` slice
# Perform a image similarity search and hit an error
```

- You select a sample that doesn't exist in the similarity index to search on (this may still raise an error if there are no valid samples selected, but that error is now also more informative)

## Changes

When subselecting samples from the lancedb table, ensure the samples being selected exist in the table.

## Testing

The above code works now

### DIFF
--- a/fiftyone/brain/internal/core/lancedb.py
+++ b/fiftyone/brain/internal/core/lancedb.py
@@ -485,10 +485,12 @@ class LanceDBSimilarityIndex(SimilarityIndex):
         df = self._table.to_pandas()
         df = df[df["id"].isin(query_ids)]
         query = np.array([v for v in df["vector"]])
+
         if query.size == 0:
             raise ValueError(
-                "Query ids %s were not found in the LanceDB index" % query_ids
+                "Query IDs %s were not found in the index" % query_ids
             )
+
         if single_query:
             query = query[0, :]
 
--- a/fiftyone/brain/internal/core/lancedb.py
+++ b/fiftyone/brain/internal/core/lancedb.py
@@ -433,8 +433,7 @@ class LanceDBSimilarityIndex(SimilarityIndex):
                 index_ids = list(self.current_sample_ids)
 
             df = table.to_pandas()
-            df.set_index("id", drop=False, inplace=True)
-            df = df.loc[df.index.isin(set(index_ids))]
+            df = df[df["id"].isin(index_ids)]
             table = self._db.create_table(
                 self.config.table_name + "_filter", df, mode="overwrite"
             )
@@ -484,8 +483,7 @@ class LanceDBSimilarityIndex(SimilarityIndex):
 
         # Query by ID(s)
         df = self._table.to_pandas()
-        df.set_index("id", drop=False, inplace=True)
-        df = df.loc[df.index.isin(set(query_ids))]
+        df = df[df["id"].isin(query_ids)]
         query = np.array([v for v in df["vector"]])
         if query.size == 0:
             raise ValueError(
--- a/fiftyone/brain/internal/core/lancedb.py
+++ b/fiftyone/brain/internal/core/lancedb.py
@@ -434,7 +434,7 @@ class LanceDBSimilarityIndex(SimilarityIndex):
 
             df = table.to_pandas()
             df.set_index("id", drop=False, inplace=True)
-            df = df.loc[index_ids]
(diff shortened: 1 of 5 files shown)


fiftyone/brain/internal/core/lancedb.py
```diff
@@ -434,7 +434,7 @@ def _kneighbors(
 
             df = table.to_pandas()
             df.set_index("id", drop=False, inplace=True)
-            df = df.loc[index_ids]
+            df = df.loc[df.index.isin(set(index_ids))]
```
kaixi-wang: why not just `df = df[df['id'].isin(index_ids)]` instead of  the set_index then filter?

ehofesmann: Yeah that should also work, good call not sure why `df.set_index()` was used to begin with

fiftyone/brain/internal/core/lancedb.py
```diff
@@ -484,9 +483,12 @@ def _parse_neighbors_query(self, query):
 
         # Query by ID(s)
         df = self._table.to_pandas()
-        df.set_index("id", drop=False, inplace=True)
-        df = df.loc[query_ids]
+        df = df[df["id"].isin(query_ids)]
         query = np.array([v for v in df["vector"]])
+        if query.size == 0:
```
brimoor, line 489: I did a quick audit of the 9 similarity backends we support and it looks like a few others also need empty query handling like this. Can you patch the `_parse_neighbors_query()` methods of those backends too while you're at it? 🤗 

Need empty query handling ⛑️⛑️⛑️
- [lancedb](https://github.com/voxel51/fiftyone-brain/blob/9d28d98b2e8e19f34d95612bac76b66f0785738c/fiftyone/brain/internal/core/lancedb.py#L490) 
- [pinecone](https://github.com/voxel51/fiftyone-brain/blob/9d28d98b2e8e19f34d95612bac76b66f0785738c/fiftyone/brain/internal/core/pinecone.py#L624)
- [qdrant](https://github.com/voxel51/fiftyone-brain/blob/9d28d98b2e8e19f34d95612bac76b66f0785738c/fiftyone/brain/internal/core/qdrant.py#L657)
- [milvus](https://github.com/voxel51/fiftyone-brain/blob/9d28d98b2e8e19f34d95612bac76b66f0785738c/fiftyone/brain/internal/core/milvus.py#L757)
- [elasticsearch](https://github.com/voxel51/fiftyone-brain/blob/9d28d98b2e8e19f34d95612bac76b66f0785738c/fiftyone/brain/internal/core/elasticsearch.py#L785)

Already have empty query handling ✅✅✅
- [sklearn](https://github.com/voxel51/fiftyone-brain/blob/9d28d98b2e8e19f34d95612bac76b66f0785738c/fiftyone/brain/internal/core/sklearn.py#L572)  
- [mongodb](https://github.com/voxel51/fiftyone-brain/blob/develop/fiftyone/brain/internal/core/mongodb.py#L678) 
- [mosaic](https://github.com/voxel51/fiftyone-brain/blob/9d28d98b2e8e19f34d95612bac76b66f0785738c/fiftyone/brain/internal/core/mosaic.py#L640) 
- [pgvector](https://github.com/voxel51/fiftyone-brain/blob/9d28d98b2e8e19f34d95612bac76b66f0785738c/fiftyone/brain/internal/core/pgvector.py#L668) 
- [redis](https://github.com/voxel51/fiftyone-brain/blob/9d28d98b2e8e19f34d95612bac76b66f0785738c/fiftyone/brain/internal/core/redis.py#L596)

ehofesmann: Done!

brimoor (approved):
LGTM

PR #287, opened 2026-04-22
Route models to video mode when applied to video collections

CharlesCNorton:
## Release Notes

- Removes duplicative model-related error handling from `fiftyone.brain`, instead deferring this to `fiftyone`'s `compute_embeddings()` method

This change allows models that support the `mode` property to be passed to methods like `compute_similarity()` and `compute_visualization()`. Currently these methods fail, even though directly passing the model to `compute_embeddings()` does work.

### DIFF
--- a/fiftyone/brain/internal/core/utils.py
+++ b/fiftyone/brain/internal/core/utils.py
@@ -787,11 +787,17 @@ def get_embeddings(
                 samples.media_type == fomm.VIDEO
                 and model.media_type == fomm.IMAGE
             ):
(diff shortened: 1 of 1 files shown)


brimoor (approved):
LGTM, assuming @CharlesCNorton is onboard with my suggestion that I've pushed to this PR: just remove the error check completely from `fiftyone.brain` and instead rely on `apply_model()` and `compute_embeddings()` to apply any necessary error-handling.

lorenzoriano (approved):

PR #289, opened 2026-05-21
Incremental embeddings visualization capabilities

prernadh:
# Rationale

Embeddings visualization on FO datasets cannot be updated when a handful of new samples are added to the dataset. The whole visualization has to be recomputed. This PR introduces functionality that allows incremental embeddings visualization brain run updates to existing brain results. 

Design doc can be found [here](https://docs.google.com/document/d/1hRRlDp2asYZpmlBt3nyfLBQiWnO1o86DKO4Q9KnnU_c/edit?tab=t.0)

## Changes

Adding a new `add_samples` method to a VisualizationResult that allows for adding samples to an existing embeddings visualization run. 

## Testing

Tested manually as well as added python tests for both sample and patch embeddings.

### DIFF
--- a/fiftyone/brain/visualization.py
+++ b/fiftyone/brain/visualization.py
@@ -38,7 +38,6 @@ logger = logging.getLogger(__name__)
 
 _DEFAULT_MODEL = "mobilenet-v2-imagenet-torch"
 _DEFAULT_BATCH_SIZE = None
-_REDUCER_BLOB_WARN_BYTES = 12_000_000
 
 
 def compute_visualization(
@@ -410,15 +409,7 @@ class VisualizationResults(fob.BrainResults):
     @property
     def reducer_blob(self):
         if self._reducer_blob is None and self._reducer is not None:
-            blob = _pickle_reducer(self._reducer)
-            if blob is not None and len(blob) > _REDUCER_BLOB_WARN_BYTES:
-                logger.warning(
-                    "Serialized visualization reducer is %d bytes, which is "
-                    "approaching MongoDB's 16 MB document limit. Persistence "
-                    "may fail.",
-                    len(blob),
-                )
-            self._reducer_blob = blob
+            self._reducer_blob = _pickle_reducer(self._reducer)
 
         return self._reducer_blob
 
@@ -1773,6 +1764,12 @@ def _expected_input_dim(reducer):
 def _strip_umap_training_data(reducer):
     reducer._raw_data = None
     if getattr(reducer, "_knn_search_index", None) is not None:
+        idx = reducer._knn_search_index
+        reducer._knn_index_params = {
+            "n_trees": idx.n_trees,
+            "n_iters": idx.n_iters,
+            "max_candidates": idx.max_candidates,
+        }
         reducer._knn_search_index = None
 
 
@@ -1789,9 +1786,7 @@ def _hydrate_umap_reducer(reducer, embeddings):
     from pynndescent import NNDescent
     from sklearn.utils import check_random_state
 
-    n = embeddings.shape[0]
-    n_trees = min(64, 5 + int(round(n**0.5 / 20.0)))
-    n_iters = max(5, int(round(np.log2(n))))
+    params = reducer._knn_index_params
 
     reducer._knn_search_index = NNDescent(
         embeddings,
@@ -1799,9 +1794,9 @@ def _hydrate_umap_reducer(reducer, embeddings):
         metric=reducer.metric,
         metric_kwds=reducer._metric_kwds,
         random_state=check_random_state(reducer.random_state),
-        n_trees=n_trees,
-        n_iters=n_iters,
-        max_candidates=60,
+        n_trees=params["n_trees"],
+        n_iters=params["n_iters"],
+        max_candidates=params["max_candidates"],
         low_memory=reducer.low_memory,
         n_jobs=reducer.n_jobs,
         verbose=reducer.verbose,
--- a/fiftyone/brain/visualization.py
+++ b/fiftyone/brain/visualization.py
@@ -1175,28 +1175,18 @@ class VisualizationResults(fob.BrainResults):
     def _filter_known(self, samples, embeddings):
         patches_field = self.config.patches_field
 
-        if patches_field is not None and self.label_ids is not None:
-            _, label_ids_path = samples._get_label_field_path(
-                patches_field, "id"
-            )
-            all_label_ids = samples.values(label_ids_path, unwind=True)
-            known_label_ids = set(self.label_ids.tolist())
-            new_label_ids = [
-                _id for _id in all_label_ids if _id not in known_label_ids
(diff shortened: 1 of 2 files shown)


fiftyone/brain/visualization.py
```diff
@@ -718,6 +751,401 @@ def remove_index(self):
...
+        embeddings, sample_ids, label_ids = fbu.get_embeddings(
+            training_view,
+            patches_field=patches_field,
+            embeddings_field=self.config.embeddings_field,
+        )
+
+        ids_returned = label_ids if patches_field is not None else sample_ids
+        order = {_id: i for i, _id in enumerate(ids_returned)}
+        idx = np.array([order[_id] for _id in training_ids if _id in order])
+        return embeddings[idx]
```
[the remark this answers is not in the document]
prernadh: addressed

fiftyone/brain/visualization.py
```diff
@@ -791,8 +1222,14 @@ def type(self):
...
 class Visualization(fob.BrainMethod):
+    supports_add_embeddings = False
+
+    def fit_reducer(self, embeddings):
+        raise NotImplementedError("subclass must implement fit_reducer()")
+
     def fit(self, embeddings):
-        raise NotImplementedError("subclass must implement fit()")
+        points, _ = self.fit_reducer(embeddings)
+        return points
```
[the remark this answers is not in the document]
prernadh: Custom Visualization backends are very unlikely and we don't know any users who have implemented their own backends, but I will add a fix for this to maintain backwards compatibility.

fiftyone/brain/visualization.py
```diff
@@ -1148,8 +1634,130 @@ def method(self):
...
+    if raw is not None:
+        return raw.shape[1]
+
+    return None
+
+
+def _strip_umap_training_data(reducer):
+    reducer._raw_data = None
+    if getattr(reducer, "_knn_search_index", None) is not None:
+        reducer._knn_search_index = None
```
lorenzoriano, line 1884: I think you should also strip `graph_` otherwise you leave a very large matrix to serialize and Mongo will complain

fiftyone/brain/visualization.py
```diff
@@ -1148,8 +1634,130 @@ def method(self):
...
+
+    reducer._knn_search_index = NNDescent(
+        embeddings,
+        n_neighbors=reducer._n_neighbors,
+        metric=reducer.metric,
+        metric_kwds=reducer._metric_kwds,
+        random_state=check_random_state(reducer.random_state),
+        n_trees=n_trees,
+        n_iters=n_iters,
+        max_candidates=60,
```
lorenzoriano: why pass the heuristics and hard coded 60? After fit you might have different values/heuristics

fiftyone/brain/visualization.py
```diff
@@ -718,6 +751,428 @@ def remove_index(self):
...
+                        ids=new_label_ids, fields=patches_field
+                    )
+                else:
+                    samples = samples.limit(0)
+        else:
+            known_samples = set(self.sample_ids.tolist())
+            all_ids = samples.values("id")
+            new_ids = [_id for _id in all_ids if _id not in known_samples]
+            if len(new_ids) != len(all_ids):
+                samples = samples.select(new_ids)
```
lorenzoriano, line 1222: when all the samples have been processed, new_ids=[], right? does `select` return an empty view in that case?

lorenzoriano (changes requested):

fiftyone/brain/visualization.py
```diff
@@ -390,6 +405,24 @@ def __init__(
 
         self.use_view(samples)
 
+    @property
+    def reducer_blob(self):
+        if self._reducer_blob is None and self._reducer is not None:
+            blob = _pickle_reducer(self._reducer)
+            if blob is not None and len(blob) > _REDUCER_BLOB_WARN_BYTES:
```
brimoor: @prernadh brain results are stored in GridFS, which does *NOT* have a 16MB limit.

However, the more data is stored on the results object, the longer it will take to call `dataset.load_brain_results()`... so there is a tradeoff in usability.

How big is the reducer in practice as a function of # embeddings?

brimoor: We could store the reducer in a *separate* GridFS blob and only load it when `add_samples()` is called. But that'd take more work and we'd need to implement it in such a way that deleting a brain result would cleanup this linked artifact.

prernadh: AH I see 😓 thanks for pointing that out! 

UMAP traditionally requires the training embeddings to be available to perform the `transform`, but I made a design choice to only allow `add_samples` when embeddings are stored on a sample field when the dimensionality reduction method is UMAP, so the reducer size really won't be that big at all and shouldn't depend much on number of embeddings used originally.

fiftyone/brain/visualization.py
prernadh: Reconstructing the `graph_` is a lot more involved than reconstructing the search_index.. also since we use GridFS, we won't hit any Mongo bottlenecks

fiftyone/brain/visualization.py
prernadh: We are not calling fit again (this is only for the `transform` call), we are just storing the original heuristics used for the search_index, umap hardcodes 60 but you are right, we should just store these values in the reducer instead of hardcoding. Will update.

fiftyone/brain/visualization.py
prernadh: Yes

fiftyone/brain/visualization.py
prernadh: Fixing the raised warning though 👍

lorenzoriano (approved):

timmydoza:
Overall this seems to work well but I did run into an interesting issue. With UMAP specifically, a crash can occur when different versions of python are used for the original brain run, and the subsequent refresh. 

For instance, UMAP runs computed from my laptop (python 3.12) crash when refreshed on a server running Python 3.11, with a numba error: `AssertionError: key already in dictionary`. I'm not sure how many of our users will be trying to use the UI to refresh a visualization that was originally created with the SDK, but this could cause much bigger problems if we ever update the version of python on the server by making all prior UMAP brain runs impossible to refresh under a new version.

Here's some python-specific context from Claude:

> The pickled reducer includes a numba-jitted distance function whose serialized code object is Python-version-specific. Unpickling succeeds, so _unpickle_reducer's version-mismatch handling never fires; the failure happens at transform time inside the JIT compile. Possible fixes: (1) the distance function is derivable from config.metric, so strip the jitted attributes before pickling and re-bind them from the local umap install on load, same spirit as the existing kNN rehydration — this makes the pickle version-stable; (2) wrap the transform call in the same recompute-guidance error unpickle failures get; (3) record the Python version on the config at fit time and check up front.

brimoor (approved):
LGTM

PR #295, opened 2026-07-10
Faster visualization results: binary storage and lighter loads

timmydoza:
# Rationale

At 500K–1M points, loading visualization results is slower than computing them. The main cause: `VisualizationResults` stores its arrays (`points`, `sample_ids`, `label_ids`) as JSON lists — a ~45 MB blob at 500K×3D whose deserialization materializes ~250 MB of transient Python objects (per server worker) to recover 12 MB of float data. Two smaller costs stack on top: every load ran a full-collection `values("id")` aggregation even when nothing needed filtering, and id filtering used a Python dict-and-loop.

## Changes

Three independent commits, by impact:

- **Store result arrays as compressed `.npy`, not JSON lists** (`b4a18d1`). The three array fields now serialize via `fou.serialize_numpy_array(ascii=True)` — zlib-compressed numpy bytes, base64-wrapped so they still ride inside the same JSON results envelope fo-core writes to GridFS today (that write path is untouched; only the array fields inside the document changed). `attributes()` excludes the arrays from the base serializer and `serialize()` re-adds them encoded. On load, `_from_dict` type-sniffs each field — a string decodes as binary, a list takes the legacy `np.array` path — so **all previously saved results load forever, no migration**, and both formats can coexist in one database indefinitely.
- **Skip `filter_ids` for root-dataset loads** (`e3b6bc1`). `use_view()` always ran `filter_ids`, which fetches every id in the collection to intersect against the index — but when the target is a root dataset, the intersection is the identity: a root dataset contains every index point by construction. The fast path skips the fetch entirely; anything with view stages goes through the existing filtering unchanged.
- **Vectorize `_parse_ids`** (`3e5a405`). Membership testing built a Python dict of id → index and looped per id. Replaced with sorted-array `searchsorted`, numpy end to end, with identical outputs — this is the same filtering the view-scoped path above still uses, so views get faster too.

Measured at 500K×3D (local M4; the ratios are the story):

| | Before | After |
|---|---|---|
| Stored blob | 45.4 MB | 17.1 MB |
| Cold deserialize (`cache=False`) | 0.44 s | 0.09 s |
| Deserialize transient memory | ~250 MB | ~0 (buffer view) |
| Cold end-to-end load, root dataset | 1.82 s | ~0.15 s |
| View-scoped load, 400K view | ~2.2 s | ~1.7 s (remainder is the id fetch itself) |

## Testing

New tests alongside the existing suite, all passing:

- `tests/test_results_serialization.py` — binary round trip, legacy-format reads, 3D, encoding contract
- `tests/test_use_view.py` — fast path, views unchanged, `test_root_dataset_is_a_snapshot`
- `tests/test_filter_ids.py` — output equivalence with the old `_parse_ids`, incl. edge cases

Also validated on a Teams ephemeral deployment at 500K and 1M points.

## Notes

- **Old versions can't read new results.** Backward compat is permanent; forward compat is not — an old server reading new-format results fails with a confusing `len() of unsized object` (we hit this during a mixed-version rollout). Needs a release-notes callout.
- **The root-dataset fast path treats results as a snapshot** of the dataset at compute time. We deliberately added no count guard: matching counts don't prove matching membership (delete one sample, add another), so it would pass silently exactly when wrong. Pinned by `test_root_dataset_is_a_snapshot`.
- **Not addressed:** `use_view()` mutates cached results in place — a latent concurrency issue with concurrent server-side view-scoping. Pre-existing, unchanged here.

### DIFF
--- a/fiftyone/brain/visualization.py
+++ b/fiftyone/brain/visualization.py
@@ -5,6 +5,7 @@ Visualization interface.
 | `voxel51.com <https://voxel51.com/>`_
 |
 """
+
 from copy import deepcopy
 import inspect
 import logging
@@ -19,6 +20,7 @@ import eta.core.utils as etau
 
 import fiftyone.brain as fb
 import fiftyone.core.brain as fob
+import fiftyone.core.dataset as fod
 import fiftyone.core.expressions as foe
 import fiftyone.core.fields as fof
 import fiftyone.core.plots as fop
@@ -527,6 +529,24 @@ class VisualizationResults(fob.BrainResults):
         Returns:
             self
         """
+        if isinstance(sample_collection, fod.Dataset):
+            # A root dataset contains every index point by construction,
+            # so skip the id aggregation that filter_ids would run just
+            # to rediscover that (measured ~1.3s at 500K points; this
+            # runs on every results load via __init__). Like the index
+            # itself, this treats the run as a snapshot: samples deleted
+            # since compute are not pruned here (view-scoped calls still
+            # prune) — refreshing the run is the supported way to sync a
+            # visualization with dataset changes
+            self._curr_view = sample_collection
+            self._curr_points = self.points
+            self._curr_sample_ids = self.sample_ids
+            self._curr_label_ids = self.label_ids
+            self._curr_keep_inds = None
+            self._curr_good_inds = None
+
+            return self
+
         sample_ids, label_ids, keep_inds, good_inds = fbu.filter_ids(
             sample_collection,
             self.sample_ids,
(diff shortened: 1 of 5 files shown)


kaixi-wang (approved):
lgtm

fiftyone/brain/visualization.py
```diff
@@ -527,6 +564,24 @@ def use_view(
         Returns:
             self
         """
+        if isinstance(sample_collection, fod.Dataset):
```
kaixi-wang, line 567: somewhat weird to be importing dataset here and ideally we wouldn't call use_view() but that might be a  more complicated fix

prernadh (approved):

fiftyone/brain/visualization.py
brimoor: This implementation is perfectly reasonable

brimoor (approved):
Very nice! Thanks for ensuring backwards compatibility! 💪

PR #296, opened 2026-07-18
Allow saving num_points in the viz 

kaixi-wang:
# Rationale

Enable storing the num_points on the config when saving the VisualizationResults so we can know the number of points without having to load the entire results object from gridfs

## Changes

- add num_points and save config when saving results

## Testing

- tested in new embeddings panel to improve initial load speed

### DIFF
--- a/fiftyone/brain/visualization.py
+++ b/fiftyone/brain/visualization.py
@@ -28,25 +28,8 @@ import fiftyone.core.utils as fou
 import fiftyone.core.validation as fov
 
 fbu = fou.lazy_import("fiftyone.brain.internal.core.utils")
-import sys, types
 
-# Block umap.parametric_umap from triggering `import tensorflow` during
-# umap/__init__.py load (causes an Abseil mutex deadlock on macOS during
-# TF's C++ static init). compute_visualization only uses umap.UMAP, never
-# ParametricUMAP, so a stub is fine.
-if "umap.parametric_umap" not in sys.modules:
-    _stub = types.ModuleType("umap.parametric_umap")
-
-    class _ParametricUMAPDisabled:
-        def __init__(self, *a, **kw):
-            raise NotImplementedError(
-                "ParametricUMAP disabled to avoid TensorFlow deadlock"
-            )
-
-    _stub.ParametricUMAP = _ParametricUMAPDisabled
-    sys.modules["umap.parametric_umap"] = _stub
-
-umap = fou.lazy_import("umap.umap_")
+umap = fou.lazy_import("umap")
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


brimoor (changes requested):
I understand the motivation here, but perhaps there is a cleaner way to achieve the desired result? How would `num_points` be used in practice? 

`config`s are intended to store static configuration-related information, but `num_points` relates to the `results`, so caching on the `config` is messy. Having to update a value on the `config` based on changes to `results` that aren't otherwise related to the `config` is non-intuitive and could easily be missed. Case and point: https://github.com/voxel51/fiftyone-brain/pull/289 updates `results` in-place. 

I assume the reason you wanted to cache on `config` is so you could access the value without having to load the entire `results` object. And unfortunately `results` are stored in GridFS, so I doubt it is possible to directly access a single key value without parsing the whole blob.

Depending on how you intend to use the cached value, I suspect there's a way to do the caching in another way. Does Embeddings 2.0 panel require creating an execution store doc for each viz? That'd be a clean place to cache data, and is consistent with how we do such things for other panels like Model Evaluation (we store cached charts).

kaixi-wang:
> I assume the reason you wanted to cache on `config` is so you could access the value without having to load the entire `results` object. And unfortunately `results` are stored in GridFS, so I doubt it is possible to directly access a single key value without parsing the whole blob.
> 
> Depending on how you intend to use the cached value, I suspect there's a way to do the caching in another way. Does Embeddings 2.0 panel require creating an execution store doc for each viz? That'd be a clean place to cache data, and is consistent with how we do such things for other panels like Model Evaluation (we store cached charts).

Yes this speeds up the first load of the new embeddings panel. In Tim's new embeddings panel, opening the panel issued a sequence of independent requests — run-info, then geometry + ids per 100k-point chunk — and every one of them loads the brain results server-side. The results cache is per-pod, and each request load-balances independently, so a single panel open can cold-load the blob on several pods, and the user stalls for the full blob-load each time a request lands on a pod that hasn't cached the run. And to pre-warm the cache, it was iterating over all brain keys to get the run-info. This doesn't scale on memory: every pod eventually loads and holds every brain result, making it more likely that all app pods crash. Only the requests that actually return points have any business loading the results. Caching num_points (and dims) on the run config at compute time makes run-info a pure metadata read that never touches the results.

timmydoza:
> The results cache is per-pod, and each request load-balances independently, so a single panel open can cold-load the blob on several pods

Just want to note that this in-memory cache, combined with the independent load balancing, is a weird paradigm for the reasons described, and it causes issues in other areas too (I ran into the same thing when investigating how signed urls are cached, for instance).

Recently, while trying the new feature to refresh an existing visualization with new samples, I saw a strange behavior that I suspect the cache causes: after a visualization is refreshed, it's possible to see different sample counts depending on the request, seemingly at random, probably based on which pod the request was routed to. Having a more stable way to access this value across all pods is something I'd be in favor of, as long as `add_samples` also updates `num_points` as part of its own save.

>  This doesn't scale on memory: every pod eventually loads and holds every brain result, making it more likely that all app pods crash. 

Great point, and this is certainly something to consider as we start to support very large embedding plots (>1M). In testing I've already seen a number of OOM crashes so I've had to bump the memory on my ephem to get around it.

PR #301, opened 2026-07-29
Actionable error for cross-environment reducer failures; opt-out for training size in get_new_ids

timmydoza:
# Rationale

Two follow-ups from the incremental refresh work in #289, both driven by the App integration (voxel51/fiftyone-teams#3425):

1. A reducer pickled under a different Python/umap/numba version unpickles cleanly but fails deep inside numba when hydrated or applied (`AssertionError: key already in dictionary: 56`), leaving users with no path forward. Root cause and repro are in [this #289 comment](https://github.com/voxel51/fiftyone-brain/pull/289#issuecomment-5073532401).
2. `get_new_ids()` unconditionally computes the training size, which unpickles the stored reducer — for UMAP that imports `umap` (~10s cold, ~200MB resident) even for callers that only need the ids. This is what froze the App's refresh form for ~10s on every cold worker.

## Changes

- Both reducer failure sites (hydration in `_prepare_reducer()` and `transform()` in `add_samples()`) now raise a recompute-required `ValueError` phrased for both UI and SDK audiences, with the original exception chained for debugging
- `get_new_ids(..., include_training_size=False)` skips the training-size computation entirely and returns `None` for it. Additive; default behavior unchanged

## Testing

- 4 new tests in `tests/test_refresh_errors.py`: the opt-out never touches the reducer, the default path still degrades gracefully on an unreadable blob, and both failure sites raise the actionable error (the transform case through the real `add_samples()` pipeline). All stub-based — no models, no zoo, `umap` never imported
- Existing lightweight suites pass unchanged

## Notes

⚠️ **Release coordination**: the App's refresh form (voxel51/fiftyone-teams#3425) now calls `get_new_ids(include_training_size=False)`, so this PR should ship in the same release as #289 — note that `release/v0.23.0` currently contains neither. If this misses the cut the form degrades gracefully (the new-sample count is omitted), but the modal-delay fix doesn't ship.

### DIFF
--- a/fiftyone/brain/visualization.py
+++ b/fiftyone/brain/visualization.py
@@ -40,6 +40,15 @@ logger = logging.getLogger(__name__)
 _DEFAULT_MODEL = "mobilenet-v2-imagenet-torch"
 _DEFAULT_BATCH_SIZE = None
 
+_REDUCER_APPLY_ERROR = (
+    "Failed to apply this visualization's stored reducer. This typically "
+    "means the run was computed in a different environment (e.g. a "
+    "different Python, umap-learn, or numba version), in which case the "
+    "stored reducer cannot be used here. Please recompute the "
+    "visualization (e.g. via compute_visualization()) to enable "
+    "incremental updates in this environment"
+)
+
 
 def compute_visualization(
     samples,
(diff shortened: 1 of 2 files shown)


fiftyone/brain/visualization.py
```diff
@@ -1051,7 +1060,12 @@ def add_samples(
                     100 * ratio,
                 )
 
-        new_points = np.asarray(self._reducer.transform(new_emb[ii, :]))
+        try:
+            new_points = np.asarray(self._reducer.transform(new_emb[ii, :]))
+        except Exception as e:
+            # same environment-mismatch failure mode as reducer hydration;
+            # see _prepare_reducer()
+            raise ValueError(_REDUCER_APPLY_ERROR) from e
```
kaixi-wang, line 1068: is it always just an environment error or could it be something else that would make the message misleading?

kaixi-wang (approved):

fiftyone/brain/visualization.py
timmydoza: Not always. Anything that fails while applying the reducer lands in this catch, so an OOM or bad embedding values would get this message too. That's why the wording says "typically means" rather than stating the cause. The original exception is also chained with `from e`, so the logs always show the real underlying error for anyone debugging. The known failure mode is a bare AssertionError from inside numba, so there isn't a more specific exception type to catch.

PR #304, opened 2026-08-05
Fixing unpickling issue when python version differs in incremental embeddings update

prernadh:
# Rationale

With UMAP, a crash can occur when different versions of python are used for the original brain run, and the subsequent refresh.

## Changes

Pickled UMAP reducers carried numba-jitted distance functions whose bytecode is Python-version-specific. Earlier we saw crashing at transform time when a run was refreshed under a different Python from the version with which the embeddings were computed. We now strip the values that were causing issues before pickling and rebind them from the local umap install on load - making the stored reducer version-portable, with bit-identical transform results.

## Testing

Manually tested

### DIFF
--- a/fiftyone/brain/visualization.py
+++ b/fiftyone/brain/visualization.py
@@ -1930,19 +1930,27 @@ def _strip_umap_training_data(reducer):
 
 
 def _rebind_umap_distance_funcs(reducer):
+    # Mirrors how UMAP.fit() binds these functions from the metric name
     import umap.distances as dist
+    from pynndescent.distances import named_distances as pynn_named_distances
 
     metric = reducer.metric
-    if not etau.is_str(metric) or metric not in dist.named_distances:
+    if etau.is_str(metric) and metric in dist.named_distances:
+        reducer._input_distance_func = dist.named_distances[metric]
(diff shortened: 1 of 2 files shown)


fiftyone/brain/visualization.py
```diff
@@ -1918,10 +1918,41 @@ def _strip_umap_training_data(reducer):
         }
         reducer._knn_search_index = None
 
+    # The jitted distance functions serialize as raw Python bytecode, which
+    # is not stable across Python versions: unpickling on a different
+    # version succeeds but numba's JIT crashes at transform time. They
+    # carry no fitted state and are fully determined by the (persisted)
+    # metric strings, so strip them here and rebind from the local umap
+    # installation on load
+    reducer._input_distance_func = None
```
kaixi-wang, line 1927: is _strip_umap_training_data called on load or on save? If the former, perhaps we should just not save and just always bind on load?

prernadh: On save, before we store the reducer.

brimoor (approved):
Approach seems reasonable to me 👍 

I didn't test, just a static review

PR #307, opened 2026-08-10
Add Python 3.13 and 3.14 support

benjaminpkane:
Extends the CI test matrix from 3.10–3.12 to 3.10–3.14 and declares the corresponding classifiers.

Notes:
- No tensorflow in the test deps, so no cp314 wheel gaps; `torch`/`torchvision` ship cp314 wheels.
- The 3.14 leg exercises `fiftyone` develop, which is gaining 3.14 support in voxel51/fiftyone#8212 — if that leg is red here, it likely needs that PR to merge first.
- Verified locally on Python 3.14.5: full install resolves and `fiftyone.brain` imports cleanly.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -65,6 +65,8 @@ jobs:
           - "3.10"
           - "3.11"
           - "3.12"
+          - "3.13"
+          - "3.14"
     steps:
       - name: Clone fiftyone-brain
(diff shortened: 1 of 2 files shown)


kevin-dimichel (approved):

kevin-dimichel:
Sneaks a lil `develop` -> `main` in there too. Carry on.

PR #311, opened 2026-08-20
Updating pgvector backend to support IVFFlat indexes

prernadh:
# Rationale

This PR adds the IVFFlat index support where we previously only supported HNSW index. Also added halfvec vector type so we can support embeddings of dim > 2000.

## Changes

Added the IVFFlat index to pgvector backend + added halfvec vector type.

## Testing

Tested with pgvector locally and added tests.

### DIFF
--- a/fiftyone/brain/internal/core/pgvector.py
+++ b/fiftyone/brain/internal/core/pgvector.py
@@ -40,6 +40,25 @@ _SUPPORTED_INDEX_TYPES = ("hnsw", "ivfflat")
 # Metrics supported by pgvector's ivfflat access method
 _IVFFLAT_SUPPORTED_METRICS = ("cosine", "dotproduct", "euclidean")
 
+# Supported vector column types for pgvector
+_SUPPORTED_VECTOR_TYPES = ("vector", "halfvec")
+
+# Metrics supported on halfvec columns (no jaccard/hamming opclasses)
+_HALFVEC_SUPPORTED_METRICS = ("cosine", "dotproduct", "euclidean", "l1")
+
+# Operator classes for halfvec columns (l1 is HNSW-only, enforced by the
+# ivfflat metric check)
+_HALFVEC_OPCLASSES = {
+    "cosine": "halfvec_cosine_ops",
+    "dotproduct": "halfvec_ip_ops",
+    "euclidean": "halfvec_l2_ops",
+    "l1": "halfvec_l1_ops",
+}
+
+# Maximum indexable dimensions per column type (applies to both hnsw and
+# ivfflat indexes)
+_MAX_INDEXABLE_DIMS = {"vector": 2000, "halfvec": 4000}
+
 # Query operators for each metric; "%" is doubled because these are
 # interpolated into parameterized psycopg2 queries
 _METRIC_OPERATORS = {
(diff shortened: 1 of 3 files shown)


fiftyone/brain/internal/core/pgvector.py
```diff
@@ -185,12 +268,33 @@ def _initialize(self):
...
+            self._cur.execute(
+                "SELECT 1 FROM pg_type WHERE typname = 'halfvec'"
+            )
+            if self._cur.fetchone() is None:
+                raise ValueError(
+                    "vector_type='halfvec' requires pgvector >= 0.7.0, but "
+                    "the 'halfvec' type was not found in the database. "
+                    "Upgrade the pgvector extension and run "
+                    "'ALTER EXTENSION vector UPDATE'"
+                )
```
AdonaiVera, line 297: This is really good, I like it. Not a blocker, just flagging that this part is missing a unit test.

fiftyone/brain/internal/core/pgvector.py
```diff
@@ -364,7 +538,7 @@ def add_to_index(
             psy_extras.execute_values(self._cur, query, data)
             self._conn.commit()
 
-        self.create_hnsw_index()
+        self.create_index()
```
AdonaiVera: Right now, adding new embeddings deletes and rebuilds the entire index. This is slow, and for IVFFlat, building with too little data can also reduce recall.

This is just an idea, not a blocker, just a suggestion, we can build the index only once. After that, let Postgres automatically add new rows to the existing index.

Something like this: 
```suggestion
        if self.config.index_name not in self._get_index_names(self.config.table_name):
                self.create_index()
```

AdonaiVera (approved):
Looks good to me @prernadh, just added some comments, nothing blocking!

Tested everything, except Azure, had to mock that call since I don't have db access there, but it works💪

fiftyone/brain/internal/core/pgvector.py
prernadh: What kind of unit test are you looking for here?

AdonaiVera: I was thinking in a mock unit test to simulate the extension setup, like CREATE EXTENSION failing on permissions, and halfvec not being there on an older pgvector. But I changed my mind, mock unit tests aren't the best fit and won't add much value here. I think we're good to go 💪

fiftyone/brain/internal/core/pgvector.py
```diff
@@ -68,8 +116,14 @@ def __init__(
...
         ssl_root_cert=None,
         work_mem="64MB",
+        maintenance_work_mem=None,
+        vector_type="vector",
+        index_type="hnsw",
         hnsw_m=16,
         hnsw_ef_construction=64,
+        hnsw_ef_search=None,
+        ivfflat_lists=100,
+        ivfflat_probes=1,
```
roboav8r: The `pgvector` README has guidance on default values here: https://github.com/pgvector/pgvector?tab=readme-ov-file#ivfflat

pgvector actually tries to warn you when the number of samples is less than list length: https://github.com/pgvector/pgvector/blob/e48241b4dcc045b18902914f668d03d1d399dfbe/src/ivfbuild.c#L466 :

```
NOTICE:  ivfflat index created with little data
DETAIL:  This will cause low recall.
HINT:    Drop the index until the table has more data.
```

So this could cause undesired results in smaller datasets if the values are fixed

brimoor (approved):
LGTM (just a static review; didn't test)

kevin-dimichel (approved):
change from develop to main looks fine to Aloha

PR #321, opened 2026-09-12
FOEPD-4662 fix(lancedb): write incrementally instead of rewriting the whole table

swheaton:
# Rationale

The LanceDB similarity backend rewrites its entire table on every `add_to_index` call — `create_table(..., mode="overwrite")`, with the rows already in the table read back and concatenated onto the new ones. Cost grows with the table rather than the batch, which puts it on the critical path for auto-indexing (FOEPD-4625).

It is also incorrect in two ways that the rewrite masks. Both reproduced against `main` at f9a55e8, adding one row and then adding the same ID again:

| | `main` | this branch |
|---|---|---|
| `overwrite=True` (the default) | ID lands in **two rows**, old vector and new | one row, new vector |
| `overwrite=False` | `UFuncTypeError` | one row, original vector |
| second add, `sample_ids` as an ndarray | `UFuncTypeError` | appends correctly |

The third row is the one that bites hardest. `fbu.get_ids` returns `np.array(...)`, and `SimilarityIndex.is_external` is `True` for LanceDB with no override, so `compute_similarity`'s "index already has the embeddings" short-circuit never fires and `add_to_index` runs on every run against a given brain key. The second such run reaches `pa_table["sample_id"].to_pylist() + sample_ids`, which broadcasts a list against an ndarray instead of concatenating, and raises. Re-running `compute_similarity` against an existing LanceDB index does not work.

## Changes

`add_to_index` upserts through `merge_insert` keyed on `id` rather than rebuilding the table, so a write costs the size of the batch. `when_not_matched_insert_all` always applies; `when_matched_update_all` applies only when `overwrite` is set, which is what makes the two `overwrite` paths above behave as documented. A duplicate ID within a single batch now raises with the offending ID, because Lance rejects a merge whose source matches one target row twice and the insert-only path would otherwise write two rows sharing an ID.

`remove_from_index` deletes by predicate instead of rewriting, and both paths batch their ID predicates at 10,000 per statement — Lance parses a predicate as one expression, so an unbounded `IN` list turns a large removal into a multi-megabyte string.

A BTree scalar index on `id` is built after the first write and rebuilt only when absent. Every merge, delete and existence check matches on `id`, and Lance scans the whole column for those matches unless it is indexed. A failure to build it warns rather than failing the add: the rows are already committed, and concurrent writers race to create it.

`_sync_table` refreshes the table handle before a write reads it. Lance pins a handle to the version it was opened at, so without the refresh another writer's rows stay invisible, an existence check misses them, and `allow_existing=False` fails to raise.

## Testing

`tests/test_lancedb.py` is new: 127 tests against real tables under `tmp_path`, skipped where `lancedb` is not installed. CI installs it, so they run there.

No version floor is declared. The suite is green at `lancedb` 0.30.2, 0.33.0, 0.38.0 and 0.39.0 — three tests that assert the id index exists skip below 0.34.0, where `create_index(config=)` arrives. The table listing is verified complete and repeat-free from 0.20.0 through 0.39.0, using `table_names` where `list_tables` does not yet exist. `ruff==0.16.0` and `black==22.3.0` clean, the version pinned in `.pre-commit-config.yaml`.

The suite was checked by mutation rather than by coverage: 52 deliberate breaks of the connector, 38 caught. Both whole-table-rewrite implementations — correct code that simply does what `main` did — are each caught by three tests, so the central claim is pinned by tests rather than only asserted in a docstring. The surviving mutations are recorded in the ticket; the ones that represent real risk are fixed in this branch's last commit.

Performance figures quoted in the docstrings are the recorded output of `tests/intensive/benchmark_lancedb.py` at 512 dimensions, a 100-row batch, on local disk. They are that harness's numbers, not measured in this PR.

## Notes

One thing out of scope: `_ensure_id_index` rebuilds only when the index is absent, never when it is stale, and nothing here calls `optimize()`. Add cost degrades against a growing unindexed tail; `optimize()` restores it.

### DIFF
--- a/fiftyone/brain/internal/core/lancedb.py
+++ b/fiftyone/brain/internal/core/lancedb.py
@@ -35,7 +35,44 @@ _SUPPORTED_METRICS = {
 # it does at 1k, where the per-call overhead dominates
 _ID_BATCH_SIZE = 10000
 
+# Page size for the deprecated `table_names()` fallback, which defaults to 10
+_TABLE_PAGE_SIZE = 1000000
+
 logger = logging.getLogger(__name__)
+
+
+def _table_names(db):
+    """Returns the names of every table in ``db``.
+
+    ``table_names()`` is paged and defaults to 10 tables, so a database holding
+    more than that reports an existing table as missing, which would strand an
+    index and let :meth:`fiftyone.brain.internal.core.utils.get_unique_name`
+    hand out a name that is already taken. ``list_tables()`` is unpaged by
+    default but is absent on older lancedb, so fall back to an explicit page
+    size there.
+    """
+    list_tables = getattr(db, "list_tables", None)
+    if list_tables is None:
+        return list(db.table_names(limit=_TABLE_PAGE_SIZE))
+
+    names = []
+    page_token = None
+    while True:
+        response = list_tables(page_token=page_token)
+
+        # Returns a paged response object, or a plain list on versions that
+        # predate paging. Materialize it so the emptiness check below tests
+        # the rows rather than the container, which may be truthy while
+        # yielding nothing
+        page = list(getattr(response, "tables", response))
+        names.extend(page)
+
+        page_token = getattr(response, "page_token", None)
+
+        # An empty page ends the walk even when a token comes back with it,
+        # which would otherwise loop forever
+        if not page_token or not page:
+            return names
 
 
 def _to_arrow_table(ids, sample_ids, embeddings):
@@ -98,7 +135,12 @@ class LanceDBSimilarityConfig(SimilarityConfig):
             provided, a new table will be created
         metric ("cosine"): the embedding distance metric to use when creating a
             new index. Supported values are ``("cosine", "euclidean")``
-        uri ("/tmp/lancedb"): the database URI to use
+        uri ("/tmp/lancedb"): the database URI to use. May be a local path or
+            an object store prefix such as ``gs://bucket/prefix``
+        storage_options (None): a dict of storage options to pass to LanceDB,
+            used to authenticate against an object store. Refer to
+            https://lancedb.github.io/lancedb/guides/storage/ for the keys each
+            store accepts
         **kwargs: keyword arguments for :class:`SimilarityConfig`
     """
 
@@ -107,6 +149,7 @@ class LanceDBSimilarityConfig(SimilarityConfig):
         table_name=None,
         metric="cosine",
         uri="/tmp/lancedb",
+        storage_options=None,
         **kwargs,
     ):
         if metric not in _SUPPORTED_METRICS:
@@ -122,6 +165,7 @@ class LanceDBSimilarityConfig(SimilarityConfig):
 
         # store privately so these aren't serialized
         self._uri = uri
+        self._storage_options = storage_options
 
     @property
     def method(self):
@@ -136,6 +180,14 @@ class LanceDBSimilarityConfig(SimilarityConfig):
         self._uri = value
 
     @property
+    def storage_options(self):
+        return self._storage_options
+
+    @storage_options.setter
+    def storage_options(self, value):
+        self._storage_options = value
+
+    @property
     def max_k(self):
         return None
 
@@ -147,8 +199,8 @@ class LanceDBSimilarityConfig(SimilarityConfig):
     def supported_aggregations(self):
         return ("mean",)
 
-    def load_credentials(self, uri=None):
-        self._load_parameters(uri=uri)
+    def load_credentials(self, uri=None, storage_options=None):
+        self._load_parameters(uri=uri, storage_options=storage_options)
 
 
 class LanceDBSimilarity(Similarity):
@@ -188,7 +240,13 @@ class LanceDBSimilarityIndex(SimilarityIndex):
 
     def _initialize(self):
         try:
-            db = lancedb.connect(self.config.uri)
+            # An object store needs credentials, which a local path does not,
+            # so only pass the options through when they were supplied
+            kwargs = {}
+            if self.config.storage_options:
+                kwargs["storage_options"] = self.config.storage_options
+
+            db = lancedb.connect(self.config.uri, **kwargs)
         except Exception as e:
             raise ValueError(
                 "Failed to connect to LanceDB backend at URI '%s'. Refer to "
@@ -196,7 +254,7 @@ class LanceDBSimilarityIndex(SimilarityIndex):
                 "information" % self.config.uri
             ) from e
 
-        table_names = db.table_names()
+        table_names = _table_names(db)
 
         if self.config.table_name is None:
             root = "fiftyone-" + fou.to_slug(self.samples._root_dataset.name)
@@ -383,7 +441,7 @@ class LanceDBSimilarityIndex(SimilarityIndex):
                     self.config.table_name, pa_table
                 )
             except Exception:
-                if self.config.table_name not in self._db.table_names():
+                if self.config.table_name not in _table_names(self._db):
                     raise
 
                 # Another writer created the table between this index opening
@@ -537,7 +595,7 @@ class LanceDBSimilarityIndex(SimilarityIndex):
             self.config.table_name,
             self.config.table_name + "_filter",
         ):
-            if tbl in self._db.table_names():
+            if tbl in _table_names(self._db):
                 self._db.drop_table(tbl)
 
         self._table = None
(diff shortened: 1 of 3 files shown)


tests/test_lancedb.py
```diff
@@ -1,23 +1,1402 @@
...
 
-Credentials, serialization, and which store a run resolves to -- none of
-which needs a database, so these run wherever the suite does. What runs
-LanceDB itself is covered in ``test_lancedb_store.py``,
-and the integration tests live in ``tests/intensive/test_similarity.py``.
+LanceDB is embedded, so these exercise a real table under ``tmp_path`` and
+need no service. They are skipped where ``lancedb`` is not installed; CI
+installs it. The tests that build an index over a dataset live in
+``tests/intensive/test_similarity.py``, and the table listing is covered
+against a real database in ``tests/test_lancedb_tables.py``.
```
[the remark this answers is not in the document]
swheaton: fixed

fiftyone/brain/internal/core/lancedb.py
```diff
@@ -237,13 +315,158 @@ def table(self):
...
+
+        try:
+            # `create_index` replaces by default, so this guard is what stops
+            # every add from rebuilding the index; it is not a
+            # micro-optimization
+            indexed_columns = [
+                table_index.columns
+                for table_index in self._table.list_indices()
+            ]
+            if ["id"] in indexed_columns:
```
kaixi-wang, line 384: This ensures there is an id index, but I don't think it ever calls optimize() to reindex, so if new rows are added, they are not in the index?

fiftyone/brain/internal/core/lancedb.py
```diff
@@ -559,7 +815,7 @@ def _parse_neighbors_query(self, query):
             single_query = False
 
         # Query by ID(s)
-        df = self._table.to_pandas()
+        df = self._rows()
```
kaixi-wang: This materialises the whole table to pull the query vectors for what is
usually a handful of IDs — and `_id_predicate` / `_get_existing_ids`, added in
this PR ~200 lines up, already do exactly this lookup as a bounded scan.

512 dims, fetching one row by ID:

| rows | via `to_pandas()` + `isin()` | via `where("id IN (...)")` | |
|---|---|---|---|
| 20,000 | 40 ms | 4.4 ms | 9x |
| 100,000 | 313 ms | 4.8 ms | 65x |
| 300,000 | 660 ms | 8.2 ms | 80x |

The gap widens linearly. At 10M × 512 this reads the full 20.5 GB of vector
payload to return one row. Cheapest large win in the diff, and the helper
already exists.

fiftyone/brain/internal/core/lancedb.py
```diff
@@ -385,7 +618,7 @@ def get_embeddings(
                     "Ignoring sample IDs when label IDs are provided"
                 )
 
-        df = self._table.to_pandas()
+        df = self._rows()
```
kaixi-wang: This materialises the whole table to pull the query vectors for what is
usually a handful of IDs — and `_id_predicate` / `_get_existing_ids`, added in
this PR ~200 lines up, already do exactly this lookup as a bounded scan.

512 dims, fetching one row by ID:

| rows | via `to_pandas()` + `isin()` | via `where("id IN (...)")` | |
|---|---|---|---|
| 20,000 | 40 ms | 4.4 ms | 9x |
| 100,000 | 313 ms | 4.8 ms | 65x |
| 300,000 | 660 ms | 8.2 ms | 80x |

The gap widens linearly. At 10M × 512 this reads the full 20.5 GB of vector
payload to return one row. Cheapest large win in the diff, and the helper
already exists.

kaixi-wang (approved):
a couple perf questions about potentially materializing all rows unnecessarily, but i'll leave it up to you if it's worth addressing now

fiftyone/brain/internal/core/lancedb.py
swheaton: Yes this is a known follow up - I will have another PR today for it

swheaton: 🤖 Claude · 💬 Answered

Right that `optimize()` is never called and that rows added after the build land in an unindexed tail — `index_stats("id_idx").num_unindexed_rows` counts them.

One correction though: they are still **found**. Lance scans the tail alongside the index, so a predicate match returns tail rows. Verified against this branch — seeded 2,000 rows, built the index, added 2 more, and `id IN ('tail-1')` returns the new row with `num_unindexed_rows == 2`. `test_finds_rows_added_after_the_index_was_built` pins it, with the reasoning that reporting them absent would make a removal silently skip them and stop `allow_existing=False` raising.

So it costs time, not correctness, and the cost is real: add degrades from 9.2 ms at 2,900 unindexed rows to 349.7 ms at 502,900, roughly doubling by 13,000. `optimize()` takes 2.7 s at 1M and restores it.

The cadence is deliberately out of scope here — it needs measuring against object storage with concurrent readers rather than picking a number, and it is tracked separately.

fiftyone/brain/internal/core/lancedb.py
swheaton: 🤖 Claude · ✅ Applied in `8f2b324`

Confirmed, and it measures worse than your numbers on my box — with the id index present and fetching one row at 512 dims:

| rows | `to_pandas()` + `isin()` | `where(id IN …)` | |
|---|---|---|---|
| 20,000 | 37.1 ms | 2.0 ms | 19x |
| 100,000 | 180.1 ms | 1.7 ms | 103x |
| 300,000 | 583.9 ms | 1.4 ms | 408x |

Both call sites now read by predicate. Added `_rows_for(ids, column="id")`, which reuses `_id_predicate` and the same `_ID_BATCH_SIZE` batching as `_get_existing_ids`:

```python
        ids = list(ids)
        if self._table is None or not ids:
            return _empty_rows()

        frames = [
            self._table.search(None)
            .where(_id_predicate(batch_ids, column=column))
            .limit(None)
            .to_pandas()
            for batch_ids in fou.iter_batches(ids, _ID_BATCH_SIZE)
        ]

        return pd.concat(frames, ignore_index=True)
```

`_parse_neighbors_query` always names IDs, so it is unconditional there. `get_embeddings` picks the column first — `sample_id` on the patches-plus-sample-ids branch, `id` otherwise — and only a request for the whole index still reads the whole table, since that is the honest read:

```python
        if lookup_ids is None:
            df = self._rows()
        else:
            df = self._rows_for(_to_id_list(lookup_ids), column=lookup_column)
```

`_id_predicate` gained a `column` argument to make the patches branch work. `TestBoundedReads` covers it by making `Table.to_pandas` raise, so a regression fails rather than just getting slower; reverting either call site fails it.

fiftyone/brain/internal/core/lancedb.py
swheaton: 🤖 Claude · ✅ Applied in `8f2b324`

Confirmed, and it measures worse than your numbers on my box — with the id index present and fetching one row at 512 dims:

| rows | `to_pandas()` + `isin()` | `where(id IN …)` | |
|---|---|---|---|
| 20,000 | 37.1 ms | 2.0 ms | 19x |
| 100,000 | 180.1 ms | 1.7 ms | 103x |
| 300,000 | 583.9 ms | 1.4 ms | 408x |

Both call sites now read by predicate. Added `_rows_for(ids, column="id")`, which reuses `_id_predicate` and the same `_ID_BATCH_SIZE` batching as `_get_existing_ids`:

```python
        ids = list(ids)
        if self._table is None or not ids:
            return _empty_rows()

        frames = [
            self._table.search(None)
            .where(_id_predicate(batch_ids, column=column))
            .limit(None)
            .to_pandas()
            for batch_ids in fou.iter_batches(ids, _ID_BATCH_SIZE)
        ]

        return pd.concat(frames, ignore_index=True)
```

`_parse_neighbors_query` always names IDs, so it is unconditional there. `get_embeddings` picks the column first — `sample_id` on the patches-plus-sample-ids branch, `id` otherwise — and only a request for the whole index still reads the whole table, since that is the honest read:

```python
        if lookup_ids is None:
            df = self._rows()
        else:
            df = self._rows_for(_to_id_list(lookup_ids), column=lookup_column)
```

`_id_predicate` gained a `column` argument to make the patches branch work. `TestBoundedReads` covers it by making `Table.to_pandas` raise, so a regression fails rather than just getting slower; reverting either call site fails it.

fiftyone/brain/internal/core/lancedb.py
swheaton: btw to_pandas() call was here long before this branch, but I should've caught it bc pandas is never the answer haha

